PR #19, opened 2019-08-08
[Bug fix] Fix skipped tests 

### LINKED ISSUE #13: stuart_ci_build skipped plugin/tests don't show up as skipped
when plugin is skipped it is not recorded as skipped.  No indication that it even exists.  This is not desired behavior.

matthewfcarlson:
Previously, when a test is meant to be skipped by the CI build it will not load the plugin, which means it doesn't show up in the JUnit XML. This fixes that. Fixes #13

### DIFF
--- a/edk2toolext/environment/plugin_manager.py
+++ b/edk2toolext/environment/plugin_manager.py
@@ -33,17 +33,12 @@ class PluginManager(object):
     # Pass tuple of Environment Descriptor dictionaries to be loaded as plugins
     #
     def SetListOfEnvironmentDescriptors(self, newlist):
-        env = shell_environment.GetBuildVars()
         failed = []
         if newlist is None:
             return []
         for a in newlist:
             b = PluginDescriptor(a)
             if(self._load(b) == 0):
-                val = env.GetValue(b.Module.upper())
-                if val and val == "skip":
-                    logging.info(f"{b.Module} turned off by environment variable")
-                    continue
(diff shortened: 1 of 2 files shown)


edk2toolext/environment/plugin_manager.py
```diff
@@ -10,7 +10,6 @@
 import os
 import imp
 import logging
-from edk2toolext.environment import shell_environment
```
corthon: By moving this to ci_build, is the implication that build_plugins can no longer be skipped on a per-platform basis?

edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -252,8 +253,17 @@ def Go(self):
                     # perhaps we should ask the validator to run on the package for this target
 
                     # Still need to see if the package decided this should be skipped
-                    if pkg_plugin_configuration is None or\
-                            "skip" in pkg_plugin_configuration and pkg_plugin_configuration["skip"]:
+                    should_skip = False
+                    if pkg_plugin_configuration is None:
+                        should_skip = True
```
corthon: By not supplying a per-package configuration, the package is skipping all plugins?

matthewfcarlson:
Fixes #13

spbrogan:
closing out.  There are numerous ways to skip a plugin and some will report and others will not.  User can get their desired behavior.

PR #21, opened 2019-08-09
[Feature] Changed how update logging works

matthewfcarlson:


edk2toolext/invocables/edk2_update.py
```diff
@@ -72,22 +72,23 @@ def GetLoggingFileName(self, loggerType):
...
+        while RetryCount < 10:  # we should put a sensible limit on the retry count for handling recursive dependencies.
             (build_env, shell_env) = self.PerformUpdate()
 
-            if not build_env_changed(build_env, build_env_old):
+            if not build_env_changed(build_env, build_env_old):  # check if the environment changed on our last update
                 break
+            # if the environment has changed, increment the retry count and notify user
+            RetryCount += 1
+            logging.log(edk2_logging.SECTION,
+                        f"Something in the environment changed. Updating the environment again. Pass #{RetryCount}")
```
spbrogan: if you are going to have a retry limit then you need to log an error/return an error/ raise an exception when the retry limit is exceeded.  If you exit with no errorcode you can't tell the difference between something broken and something working.

PR #26, opened 2019-08-10
Add API to VarDict to alllow override

spbrogan:
VarDict has a unique feature in that it can block override of a value.  This is heavily used in the tool and build flow to make sure that settings with high priority are not overridden later in the process by something with lower priority.  Also it is critical for stability that once a value is set and used that it doesn't change.  

But in some cases for specific usage models it might be needed.  The API makes it possible.  
One such example case is in adapting support for Host Based Unit tests and wanted to only compile Host Tests that will run on the host afterwards.


spbrogan:
Resolves issue #27 and provides new API to support required behavior.

PR #33, opened 2019-08-28
Added helpful message when we use the wrong invokable type on a given py file

### LINKED ISSUE #23: Stuart doesn't give a good error when trying to use the wrong invokable
When doing a stuart_build on a CISettings.py file, it just says 'NoneType' object is not callable - which is fairly cryptic. Checking for this common mistake would cut down on a lot of developer confusion.

matthewfcarlson:
Fixes #23

### DIFF
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -15,6 +15,7 @@ from edk2toolext.environment import shell_environment
 from edk2toollib.utility_functions import GetHostInfo
 from edk2toollib.utility_functions import locate_class_in_module
 from edk2toollib.utility_functions import import_module_by_file_name
+from edk2toollib.utility_functions import is_instance
 from edk2toolext.base_abstract_invocable import BaseAbstractInvocable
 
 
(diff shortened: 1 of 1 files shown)


matthewfcarlson:
@spbrogan this is ready :)

edk2toolext/edk2_invocable.py
```diff
@@ -104,18 +106,42 @@ def ParseCommandLineOptions(self):
...
                 self.AddCommandLineOptions(settingsParserObj)
+                Module = self.PlatformModule
+                module_contents = dir(Module)
+                # Filter through the Module, we're only looking for classes.
+                classList = [getattr(Module, obj) for obj in module_contents if inspect.isclass(getattr(Module, obj))]
+                # Get the name of all the classes
+                classNameList = [obj.__name__ for obj in classList if is_instance(obj)]
+                # TODO improve is_instance as it catches imports as well as declared classes
+                imported_classes = ", ".join(classNameList)  # Join the classes together
+                print(f"The module you imported contains {imported_classes}")
```
spbrogan: can these print statements not be done using the logging module?  logging.critical seems like it would be better here unless this happens before logging is up?

matthewfcarlson: We don't have logging setup up yet, so we can't do logging.

matthewfcarlson: We need the platform settings to get the loglevel and settings

matthewfcarlson:
@spbrogan ready to go

PR #35, opened 2019-08-29
Explicit Package Paths

### LINKED ISSUE #31: stuart_ci_build should  not put all dependency in the PackagesPath

matthewfcarlson:
Fixes #31

### DIFF
--- a/edk2toolext/invocables/edk2_ci_build.py
+++ b/edk2toolext/invocables/edk2_ci_build.py
@@ -132,8 +132,10 @@ class Edk2CiBuild(Edk2Invocable):
         pplist = self.PlatformSettings.GetPackagesPath() if self.PlatformSettings.GetPackagesPath() else []
 
         # Check Dependencies for Repo
+        # Generally we want the dependencies to be on the package path but not explictly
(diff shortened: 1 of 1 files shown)


edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -132,8 +132,11 @@ def Go(self):
         pplist = self.PlatformSettings.GetPackagesPath() if self.PlatformSettings.GetPackagesPath() else []
 
         # Check Dependencies for Repo
+        # Generally we want the dependencies to be on the package path but not explictly
         for dependency in self.PlatformSettings.GetDependencies():
```
spbrogan: why do you want to log a warning. ?

matthewfcarlson: I was thinking that if a dependency isn't on the packages path, it can cause rather strange errors. This was just to throw a warning to let a user know it's not on the path so it's a little easier to track down the error

spbrogan:
i don't like this.  I think it is dictating too much based on assumptions.  I would be interested others opinions @tianocore/edk-ii-tool-maintainers ??

PR #36, opened 2019-08-29
[Bug] Added check for non-zero return code on git clones

### LINKED ISSUE #30: Stuart_setup doesn't give a good error when unable to clone a specific commit

matthewfcarlson:
Fixes #30


spbrogan:
can we just dump all this git code in setup and switch back to our git libraries.  We should not be creating more git wrappers.  We already have edk2_git and repo resolver.  I think a refactor is needed.

matthewfcarlson:
@spbrogan let's complete this until we redo the git code

edk2toolext/invocables/edk2_setup.py
```diff
@@ -177,8 +177,11 @@ def Go(self):
                         if self.omnicache_path is not None:
                             cmd_string += " --reference " + self.omnicache_path
                         cmd_string += " " + required_repo
-                        RunCmd('git', cmd_string, workingdir=workspace_path,
-                               logging_level=logging.DEBUG, raise_exception_on_nonzero=True)
+                        ret = RunCmd('git', cmd_string, workingdir=workspace_path,
+                                     logging_level=logging.DEBUG, raise_exception_on_nonzero=False)
+                        if ret != 0:
+                            logging.error("We failed to fetch " + required_repo)
+                            raise ValueError("Unable to checkout repo due to error")
```
corthon: We should change this to RuntimeError so that it's caught by the except and doesn't create an alternate error path.

matthewfcarlson: Agreed. Fixing.

PR #42, opened 2019-08-30
[Bug] misspelling and non-empty path on VsWhere

matthewfcarlson:
Check to make sure we actually have a path when we get a result back from VsWhere

### DIFF
--- a/edk2toolext/environment/conf_mgmt.py
+++ b/edk2toolext/environment/conf_mgmt.py
@@ -194,7 +194,7 @@ class ConfMgmt():
             if(path is None):
                 # Not specified...find latest
                 (rc, path) = FindWithVsWhere(vs_version=vsversion)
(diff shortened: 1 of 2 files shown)


edk2toolext/environment/conf_mgmt.py
```diff
@@ -194,7 +194,7 @@ def GetVsInstallPath(vsversion, varname):
             if(path is None):
                 # Not specified...find latest
                 (rc, path) = FindWithVsWhere(vs_version=vsversion)
-                if rc == 0 and path is not None:
+                if rc == 0 and path is not None and len(path) > 0:
```
spbrogan: to make better should we call os.path.exists()
https://docs.python.org/3.7/library/os.path.html#os.path.exists

matthewfcarlson: fixed

Issue #47, opened 2019-09-09
Setup, Update, and Build SettingsManager can't optimize based on user supplied CLI options because there is no sharing of that back to settings manager


spbrogan:
In trying to optimize the ci build time more has been identified that can be done if the SettingsManagers know more about what is going to actually build.  This means things like individual packages, architectures, and targets.   The SettingsManagers must be notified of the CLI settings prior to providing the active scopes or required repos.

A few optimizations that should be supported:
* If a run is not going to build cryptopkg then it might not need the OpenSSL submodule. 
  * This means that stuart setup, ci setup, and update needs to optionally be told what package is going to be built so that submodules and dependencies could be managed. 
* A run might only want to compile for DEBUG or RELEASE.
  * The Target should be settable on the cli.
* A run might only want to compile/test for a given arch
  * Arch should be settable on the cli

spbrogan:
resolved by PR #48

PR #48, opened 2019-09-09
Refactor allow actual settings

spbrogan:


spbrogan:
This attempts to resolve issue #47

spbrogan:
This is a breaking change.  
This has been tested on a platform build and ci builds.

spbrogan:
this also resolves #31

matthewfcarlson:
Fixed #31

matthewfcarlson:
@spbrogan is there an example of a CI file that uses this? I'm trying to evaluate the flow and I'd like to see it

edk2toolext/base_abstract_invocable.py
```diff
@@ -45,6 +45,10 @@ def GetLoggingFolderRelativeToRoot(self):
         ''' Return a path to folder for log files '''
         raise NotImplementedError()
 
+    def InputParametersConfigured(self):
```
matthewfcarlson: Maybe we should add Hook or Callback to make it a little more clear

edk2toolext/base_abstract_invocable.py
```diff
@@ -45,6 +45,10 @@ def GetLoggingFolderRelativeToRoot(self):
...
+        ''' Do any necessary work once all the input parameters are configured. '''
```
matthewfcarlson: This function is called once all the input parameters are collected and can be used to initialize environment...

edk2toolext/base_abstract_invocable.py
```diff
@@ -97,6 +101,7 @@ def Invoke(self):
         ''' Main process function.  Should not need to be overwritten '''
 
         self.ParseCommandLineOptions()
+        self.InputParametersConfigured()
```
matthewfcarlson: Should we setup logging before we call this function? Maybe we should be more clear about what functions have logging setup. Right now it's only ParseCommandLineOptions

edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -75,6 +80,33 @@ def GetPluginSettings(self):
         '''  Implement in subclass to pass dictionary of settings for individual plugins '''
         return {}
```
matthewfcarlson: Oh we do have a plugin settings function

edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -75,6 +80,33 @@ def GetPluginSettings(self):
...
+
+    def SetToArchitecture(self, list_of_requested_architectures):
+        ''' Confirm the requests architecture list is valid and configure SettingsManager
+        to run only the requested architectures.
+
+        Raise Exception if a list_of_requested_architectures is not supported
+        '''
+        pass
+
+    def SetToTarget(self, list_of_requested_target):
```
matthewfcarlson: I'm not super clear on what the SetToArchitecture and the SetTo* stuff does/where it's called/etc

matthewfcarlson: Ah I see it.

matthewfcarlson: So the default implementation is to do nothing- is this desired behavior? Or would we want to throw a not implemented exception

edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -193,12 +244,8 @@ def Go(self):
         logging.log(edk2_logging.SECTION, "Loading plugins")
 
         pluginList = self.plugin_manager.GetPluginsOfClass(ICiBuildPlugin)
-        if len(self.packageList) == 0:
-            self.packageList.extend(self.PlatformSettings.GetPackages())
-
-        for pkgToRunOn in self.packageList:
 
-            # TODO: figure out if this package is a valid package
```
matthewfcarlson: We still need to figure out if this is a valid package or not

edk2toolext/invocables/edk2_setup.py
```diff
@@ -48,6 +48,49 @@ def GetLoggingLevel(self, loggerType):
         '''
         pass
 
+    # ####################################################################################### #
+    #                           Supported Values and Defaults                                 #
+    # ####################################################################################### #
+    def GetPackagesSupported(self):
```
matthewfcarlson: I think it could be worthwhile to have base settings classes where we include these functions like GetArchitecturesSupported

edk2toolext/invocables/edk2_update.py
```diff
@@ -65,12 +108,58 @@ def GetVerifyCheckRequired(self):
...
+        '''  Providing UpdateSettingsManger '''
         return UpdateSettingsManager
 
     def GetLoggingFileName(self, loggerType):
         return "UPDATE_LOG"
 
+    def AddCommandLineOptions(self, parserObj):
+        ''' adds command line options to the argparser '''
+        # This will parse the packages that we are going to update
+        parserObj.add_argument('-p', '--pkg', '--pkg-dir', dest='packageList', type=str,
```
matthewfcarlson: Additionally, we could move this out to some sort of common file for Target/Package/Arch and pass in the ParserObj to get these added

edk2toolext/invocables/edk2_ci_build.py
matthewfcarlson: SetSettingsToTarget might be better

matthewfcarlson: Or should we expect the CI settings to read from the environment?

edk2toolext/invocables/edk2_setup.py
spbrogan: can't really because these command line options are not defined except in the final instance

edk2toolext/invocables/edk2_ci_build.py
spbrogan: This is handled in the Set function.  It raises exception if unknown or unsupported

edk2toolext/base_abstract_invocable.py
spbrogan: agree

edk2toolext/base_abstract_invocable.py
spbrogan: agree.  Moved to after logging

edk2toolext/invocables/edk2_ci_build.py
spbrogan: fixed func names.  
Default is to do nothing and not raise.  I think that is choice of implementation
And i think for these 3 settings they are special

edk2toolext/invocables/edk2_update.py
spbrogan: created a new intermediate base class.  Looks better. thanks for the feedback

PR #50, opened 2019-09-09
check to make sure we aren't putting in new stripped lines

### LINKED ISSUE #46: BuildConf.config syntax with a trailing comment will not work
See
https://github.com/tianocore/edk2-pytool-extensions/blob/master/edk2toolext/edk2_invocable.py#L158

If the config file has
test=goingToFail #comment here

This will incorrectly try to set test = goingToFail #comment here
whereas it should set test = goingToFail

### LINKED ISSUE #49: Build.Conf file throw errors if two empty lines are in it
Sample file provided
```
LaunchBuildLogProgram = "C:\Program Files\Sublime Text 3\subl.exe"  #if you remove this it will launch with notepad by default
LaunchLogOnSuccess = False  #don't want to see build log on success  # set to true if you want to see it
LaunchLogOnError = False  #don't want to see build log on error # set to true if you want to see it
BUILDREPORT_TYPES=PCD DEPEX FLASH BUILD_FLAGS LIBRARY


BUILDREPORTING=TRUE
```

matthewfcarlson:
Fixes #46 and fixes #49

### DIFF
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -155,8 +155,11 @@ Key=value will get passed to build process for given build type)'''
             with open(BuildConfig) as file:
                 for line in file:
                     stripped_line = line.strip()
-                    if not stripped_line.startswith("#"):
-                        unknown_args.append(line)
+                    if stripped_line.startswith("#"):
+                        continue
+                    if len(stripped_line) == 0:
+                        continue
+                    unknown_args.append(stripped_line)
 
         i = 0
         while i < len(unknown_args):


edk2toolext/edk2_invocable.py
```diff
@@ -151,12 +152,16 @@ def ParseCommandLineOptions(self):
...
 
+        comment_re = re.compile(r"#.+$")
+
         if os.path.isfile(BuildConfig):
             with open(BuildConfig) as file:
                 for line in file:
                     stripped_line = line.strip()
-                    if not stripped_line.startswith("#"):
-                        unknown_args.append(line)
+                    stripped_line = comment_re.sub("", stripped_line)
```
spbrogan: I am not a regex power user but the python partition function could work and would require no additional imports and fewer lines.  

Basically 
```python
stripped_line = stripped_line.partition("#")[0]
```

https://docs.python.org/3/library/stdtypes.html?highlight=partition#str.partition

edk2toolext/edk2_invocable.py
```diff
@@ -151,12 +152,16 @@ def ParseCommandLineOptions(self):
...
+                    if len(stripped_line) == 0:
+                        continue
+                    unknown_args.append(stripped_line)
```
spbrogan: I don't really like the BuildConfig values getting added to the unknown_args and then added to the BuildVars with the comment "From CmdLine".  

Can we update logic to correct the string so that cmd line variables and BuildConfig values are treated differently.

edk2toolext/edk2_invocable.py
matthewfcarlson: Good point

edk2toolext/edk2_invocable.py
matthewfcarlson: Logic has been updated

PR #54, opened 2019-09-13
[Bug] added check for bad path when searching for VsWhere

matthewfcarlson:


edk2toolext/environment/conf_mgmt.py
```diff
@@ -204,6 +204,9 @@ def GetVsInstallPath(vsversion, varname):
         def GetVcVersion(path, varname):
             # check if already specified
             vc_ver = shell_environment.GetEnvironment().get_shell_var(varname)
+            if (path is None):
+                self.Logger.critical("Failed to find VC tools.  Might need to check for VS install")
```
spbrogan: can this log message be unique...so it is easier to root cause. 
Also i think if path is none then basically it failed to Visual Studio not the Visual C compiler.  So you could just change to "Failed to find VS tools"

matthewfcarlson: Great point about making a unique error message

PR #58, opened 2019-09-26
[Bug] fix architecture setup in ci build

matthewfcarlson:


edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -132,7 +132,12 @@ def Go(self):
             pc = '"' + pc + '"'
         shell_env.set_shell_var("PYTHON_COMMAND", pc)
 
-        env.SetValue("TARGET_ARCH", " ".join(self.requested_architecture_list), "from edk2 ci build.py")
+        try:
+            arches = self.PlatformSettings.ActualTargets
```
spbrogan: not defined member of interface.  stuart can only use what is defined in settingsmanager

PR #61, opened 2019-09-26
VersionAggregator improvements (pip version and python version)

### LINKED ISSUE #59: Missing information from version report
1. pip versions
2. Python version

matthewfcarlson:
fixes #59

### DIFF
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -38,8 +38,8 @@ class Edk2Invocable(BaseAbstractInvocable):
         # go through all installed pip versions
         for package in pip_packages:
             version = pkg_resources.get_distribution(package).version
-            logging.info("{0} version: {1}".format(package, version))
-            ver_agg.ReportVersion(package, version, version_aggregator.VersionTypes.TOOL)
+            logging.info("{0} version: {1}".format(package.project_name, version))
+            ver_agg.ReportVersion(package.project_name, version, version_aggregator.VersionTypes.TOOL)
 
     def GetWorkspaceRoot(self):
         try:
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -12,15 +12,33 @@ import os
 import sys
 import logging
 import inspect
+import pkg_resources
 import argparse
 from edk2toolext.environment import shell_environment
 from edk2toollib.utility_functions import GetHostInfo
+from edk2toolext.environment import version_aggregator
 from edk2toollib.utility_functions import locate_class_in_module
 from edk2toollib.utility_functions import import_module_by_file_name
 from edk2toolext.base_abstract_invocable import BaseAbstractInvocable
 
 
 class Edk2Invocable(BaseAbstractInvocable):
+
+    #
+    # Pass in a list of pip package names and they will be printed as well as
+    # reported to the global version_aggregator
+    @classmethod
+    def _collect_python_pip_info(cls):
+        # Get the current python version
+        cur_py = "%d.%d.%d" % sys.version_info[:3]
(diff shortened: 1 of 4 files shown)


matthewfcarlson:
It looks like we're not output the version information to a file anywhere unfortunately. Not sure if that's needed or not.

matthewfcarlson:
I can verify that it puts it out into the build log

spbrogan:
we need to solve this for platform builds and ci builds.  

It would be ideal if this was resolved for update and setup steps too as this information can be useful when debugging issues. 

Thoughts?

matthewfcarlson:
@spbrogan I agree. I think the plugin that actually extracts the version aggregation is a uefibuild plugin, so we'll need to move it into something that is common across everything. I'm not sure we have a category for this.

spbrogan:
I would suggest we don't worry about the output yet since it does get into the standard log.  For CI and platform build the output will also be saved to log file by the existing plugin.

edk2toolext/edk2_invocable.py
```diff
@@ -12,16 +12,35 @@
...
+        cur_py = "%d.%d.%d" % sys.version_info[:3]
+        ver_agg = version_aggregator.GetVersionAggregator()
+        ver_agg.ReportVersion("Python", cur_py, version_aggregator.VersionTypes.TOOL)
+        # Get a list of all the packages currently installed in pip
+        pip_packages = [p for p in pkg_resources.working_set]
+        # go through all installed pip versions
+        for package in pip_packages:
+            version = pkg_resources.get_distribution(package).version
+            logging.info("{0} version: {1}".format(package.project_name, version))
+            ver_agg.ReportVersion(package.project_name, version, version_aggregator.VersionTypes.TOOL)
```
spbrogan: Maybe a different type for pip packages since they aren't really tools

edk2toolext/edk2_invocable.py
```diff
@@ -12,16 +12,35 @@
...
 from edk2toollib.utility_functions import GetHostInfo
+from edk2toolext.environment import version_aggregator
 from edk2toollib.utility_functions import locate_class_in_module
 from edk2toollib.utility_functions import import_module_by_file_name
 from edk2toolext.base_abstract_invocable import BaseAbstractInvocable
 
 
 class Edk2Invocable(BaseAbstractInvocable):
 
+    #
```
spbrogan: incorrect function comment

edk2toolext/environment/version_aggregator.py
```diff
@@ -46,6 +46,12 @@ def ReportVersion(self, key, value, versionType):
         }
         self._logger.debug("version_aggregator logging version: {0}".format(str(self.Versions[key])))
 
+    def Print(self):
+        """ Prints out the current information from the version aggregator """
+        for version_key in self.Versions:
+            version = self.Version[version_key]
+            print(f"{version['type']} - {version['name']}: {version['version']}")
```
spbrogan: Are these objects? should we instead provide a str function so that every place this is desired don't have to write the same format string?

edk2toolext/edk2_invocable.py
```diff
@@ -12,16 +12,35 @@
...
+    # Pass in a list of pip package names and they will be printed as well as
+    # reported to the global version_aggregator
+    @classmethod
+    def _collect_python_pip_info(cls):
```
spbrogan: I don't think a class function should start with _ since that is an indicator of private and by definition a class function is not private

edk2toolext/environment/version_aggregator.py
matthewfcarlson: They're dictionaries :/

edk2toolext/edk2_invocable.py
matthewfcarlson: Fair

edk2toolext/edk2_invocable.py
matthewfcarlson: Good spot

Issue #65, opened 2019-10-02
web dependency sha256 value is case sensitive


spbrogan:
comparison or sha256 hex string expects lower case.

 Make case insensitive and unit test as there is no coverage on the sha256 comparison code

PR #66, opened 2019-10-02
Bug: Fix web dependency hash compare

spbrogan:
Add test coverage for sha256 on web dependency
Resolve bug of case sensitivity 

Issue #65


spbrogan:
fixes #65

PR #70, opened 2019-10-02
added naming for buildlogs

matthewfcarlson:


matthewfcarlson:
Yes- hadn't thought of that. Would you want to add a new method to the settings, get_logging_filename?

PR #74, opened 2019-10-09
Submodule recursion support to optimize initialization process

spbrogan:
Breaking change to edk2_setup.py settings API

### DIFF
--- a/edk2toolext/invocables/edk2_ci_build.py
+++ b/edk2toolext/invocables/edk2_ci_build.py
@@ -26,13 +26,27 @@ class CiBuildSettingsManager(MultiPkgAwareSettingsInterface):
     ''' Platform settings will be accessed through this implementation. '''
 
     def GetActiveScopes(self):
-        ''' get scope '''
+        ''' return tuple containing scopes that should be active for this process '''
         raise NotImplementedError()
 
     def GetDependencies(self):
+        ''' Return Git Repository Dependendencies
+        TODO:::how used in ci_build
+
+        Return an iterable of dictionary objects with the following fields
+        {
+            Path: <required> Workspace relative path
+            Url: <required> Url of git repo
+            Commit: <optional> Commit to checkout of repo
+            Branch: <optional> Branch to checkout (will checkout most recent commit in branch)
+            Full: <optional> Boolean to do shallow or Full checkout.  (default is False)
+            ReferencePath: <optional> Workspace relative path to git repo to use as "reference"
+        }
+        '''
         pass
 
     def GetPackagesPath(self):
+        ''' Return a list of workspace relative paths that should be mapped as edk2 PackagesPath '''
         pass
 
     # ####################################################################################### #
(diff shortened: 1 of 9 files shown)


edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -26,13 +26,11 @@ class CiBuildSettingsManager(MultiPkgAwareSettingsInterface):
     ''' Platform settings will be accessed through this implementation. '''
 
     def GetActiveScopes(self):
-        ''' get scope '''
+        ''' return tuple containing scopes that should be active for this process '''
```
matthewfcarlson: maybe not tuple but iterable

edk2toolext/invocables/edk2_setup.py
```diff
@@ -113,8 +132,8 @@ def Go(self):
                 # Clean any submodule repos.
                 if required_repos:
```
matthewfcarlson: maybe these variables should be renamed (vs code makes this really easy with refactor > rename or f2). required_submodule instead of required_repo

edk2toolext/invocables/edk2_setup.py
```diff
@@ -130,35 +149,37 @@ def Go(self):
...
             # Git Repos: STEP 1 --------------------------------------
             # Make sure that the repos are all synced.
-            try:
-                edk2_logging.log_progress("## Syncing Git repositories: %s..." % ", ".join(required_repos))
-                RunCmd("git", 'submodule sync -- ' + " ".join(required_repos),
-                       workingdir=workspace_path, logging_level=logging.DEBUG, raise_exception_on_nonzero=True)
+            for required_repo in required_repos:
+                try:
+                    edk2_logging.log_progress(f"## Syncing Git repositories: {required_repo.path}...")
+                    RunCmd("git", f'submodule sync -- {required_repo.path}',
```
matthewfcarlson: Will this be a perf hit? I think there's a large startup cost of invoking git. You can use a python map expression to create the list of needed for the old code. [x.path for x in required_repos]

edk2toolext/invocables/edk2_setup.py
```diff
@@ -173,10 +194,13 @@ def Go(self):
                     # If we're not skipping, grab it.
                     if not skip_repo or self.force_it:
                         logging.info("## Fetching repo.")
-                        cmd_string = "submodule update --init --recursive --progress"
+                        cmd_string = "submodule update --init"
```
matthewfcarlson: A pattern that we use is creating a list and appending to the list conditionally and then joining it later. Not sure if you want to use that here.

edk2toolext/invocables/edk2_ci_build.py
spbrogan: we add to it and we add a tuple so i think it must be a tuple

edk2toolext/invocables/edk2_setup.py
spbrogan: i don't think it is worth it here.

edk2toolext/invocables/edk2_setup.py
spbrogan: agree

edk2toolext/invocables/edk2_setup.py
spbrogan: ok

PR #84, opened 2019-10-11
Ci/improve coverage

matthewfcarlson:
Takes us from 50% to 63%


edk2toolext/tests/test_edk2_plat_build.py
```diff
@@ -0,0 +1,123 @@
...
+import unittest
+from edk2toolext.invocables.edk2_platform_build import Edk2PlatformBuild
+import tempfile
+import sys
+import os
+import logging
+from importlib import reload
+from edk2toolext.environment import shell_environment
+
+class_info = '''
```
corthon: Rather than take this approach to spontaneous file creation (since this has to be done for almost every test), could we instead move all these tests into a sub-directory of tests/ and include test files alongside them?

matthewfcarlson: Hmmm- it's an interesting idea. Have a minimum viable uefi tree rather than dynamically creating one.

matthewfcarlson: Alright- tried it out, let me know what you think?

edk2toolext/tests/test_edk2_ci_build.py
```diff
@@ -0,0 +1,58 @@
...
+from importlib import reload
+from edk2toolext.environment import shell_environment
+
+
+class TestEdk2CiBuild(unittest.TestCase):
+
+    minimalTree = os.path.join(os.path.dirname(__file__), "minimal_uefi_tree")
+
+    def setUp(self):
+        TestEdk2CiBuild.restart_logging()
```
corthon: What's the significance of all these restarts?

matthewfcarlson: Comment added

PR #88, opened 2019-10-15
Bug fix: logic to detect public file dependency in pr eval

spbrogan:
Logic bug as it did not check if the file was public before triggering dependencies to build.


spbrogan:
fixes #90

PR #91, opened 2019-10-15
Update nuget version to 5.3

matthewfcarlson:
Update nuget version to 5.3 from 5.1, which might help with some linux credential provider issues


spbrogan:
Have you tested on windows?  does this work?  do auth feeds work?  are we confident there are no regressions?

matthewfcarlson:
So thanks to the unit tests we know that most of the nuget functionality works but I will test authenticated feeds in my build tree today

spbrogan:
@matthewfcarlson  waiting on your authenticated feed test

matthewfcarlson:
I tested it on windows on authenticated feeds. I didn't do authenticated feeds on linux, but it looked like it worked well getting a regular feed for linux.

PR #97, opened 2019-10-22
Platform Build Settings manager can now return uefibuilder class

matthewfcarlson:
When dealing with multiple classes, the Platform Build Settings Manager can return the Uefi Builder class that should be used

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -91,6 +91,7 @@
         "CISETUP",
         "isclass",
         "DSCs",
-        "levelno"
+        "levelno",
+        "uefibuilder"
     ]
 }
--- a/.cspell.json
+++ b/.cspell.json
@@ -91,7 +91,6 @@
         "CISETUP",
         "isclass",
         "DSCs",
-        "levelno",
-        "uefibuilder"
+        "levelno"
     ]
 }
--- a/.cspell.json
+++ b/.cspell.json
@@ -91,6 +91,10 @@
         "CISETUP",
         "isclass",
         "DSCs",
-        "levelno"
+        "levelno",
+        "pyopenssl",
+        "esrt",
+        "noqa",
+        "pymodule"
     ]
 }
--- a/edk2toolext/invocables/edk2_platform_build.py
+++ b/edk2toolext/invocables/edk2_platform_build.py
@@ -37,6 +37,10 @@ class BuildSettingsManager():
         ''' Get the name of the repo, platform, or product being build '''
         pass
 
+    def GetUefiBuilderClass(self):
+        ''' Returns the platform builder class that we want to use '''
+        pass
+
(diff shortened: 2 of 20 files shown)


.cspell.json
```diff
@@ -91,6 +91,7 @@
         "CISETUP",
         "isclass",
         "DSCs",
-        "levelno"
+        "levelno",
+        "uefibuilder"
```
spbrogan: since your function is GetUefiBulderClass which is handled by spelling because of capitalization i would prefer that you use UefiBuilder in your comment and therefore this isn't needed.

edk2toolext/invocables/edk2_platform_build.py
```diff
@@ -37,6 +37,10 @@ def GetName(self):
         ''' Get the name of the repo, platform, or product being build '''
         pass
 
+    def GetUefiBuilderClass(self):
+        ''' Returns the platform builder class that we want to use '''
```
corthon: Class should be a string? Can you clarify the comment?

edk2toolext/invocables/edk2_platform_build.py
```diff
@@ -60,10 +64,17 @@ class Edk2PlatformBuild(Edk2Invocable):
 
     def AddCommandLineOptions(self, parserObj):
         ''' adds command line options to the argparser '''
+    # check to see if the settings manager has uefibuilder
+        uefi_build_class = self.PlatformSettings.GetUefiBuilderClass()
 
         # PlatformSettings could also be a subclass of UefiBuilder, who knows!
         if isinstance(self.PlatformSettings, UefiBuilder):
             self.PlatformBuilder = self.PlatformSettings
+        elif uefi_build_class:
```
corthon: This expects that a non-implementation would return None, right? Just confirming I understand the flow.

matthewfcarlson: Yes. Non-implementations would return None

edk2toolext/invocables/edk2_platform_build.py
matthewfcarlson: This allows a setting file to return the class, not an object that we want to create to use as a uefi builder

.cspell.json
matthewfcarlson: Will remove

matthewfcarlson:
@spbrogan The reason this is was created was to be specific about the platform builder that we want to use. Right now what we have works fine since we just import the platformbuildworker.py and it works by virtue of being imported. Pip, if turned on, would complain because it's a useless import that doesn't get used elsewhere that it can see. So we create a method that returns the thing that we import.

matthewfcarlson:
I'm okay with this PR being closed if it is decided that this isn't wanted

PR #100, opened 2019-10-29
Add CLI and helper modules to assist in capsule creation and signing.

corthon:
Needs an extra CLI option for CAT file signing, but this is mostly there.


requirements.txt
```diff
@@ -2,3 +2,4 @@ pytest
 pytest-html
 pytest-cov
 flake8
+pyopenssl
```
corthon: Should this be removed?

setup.py
```diff
@@ -70,11 +70,13 @@ def run(self):
...
                             'omnicache=edk2toolext.omnicache:main',
-                            'nuget-publish=edk2toolext.nuget_publishing:go']
+                            'nuget-publish=edk2toolext.nuget_publishing:go',
+                            'edk2_capsule_tool=edk2toolext.capsule.capsule_tool:main']
     },
     install_requires=[
         'pyyaml',
-        'edk2-pytool-library>=0.9.1'
+        'edk2-pytool-library>=0.10.3'
+        # Update with pyopenssl?
```
corthon: Was unsure about prereqs for the optional module.

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,162 @@
...
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+# IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+# INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+# OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+# ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
spbrogan: wrong license text

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,162 @@
...
+    # Start building the capsule as we go.
+    # Create the FMP Payload and set all the necessary options.
+    fmp_payload_header = FmpPayloadHeaderClass()
+    fmp_payload_header.FwVersion = int(capsule_options['fw_version'], 16)
+    fmp_payload_header.LowestSupportedVersion = int(capsule_options['lsv_version'], 16)
+    fmp_payload_header.Payload = capsule_data
+
+    # Create the auth header and get ready to sign the data.
+    fmp_auth_header = FmpAuthHeaderClass()
+    # TODO: Allow caller to pass the monotonic count in capsule_options.
```
spbrogan: I wouldn't suggest it.  
To date we have said monotonic count is a broken feature and we don't allow usage.  I would add comment why monotonic count is not a good feature and don't allow caller to set it.

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Sign the data and assign it to the cert data.
+    # TODO: Determine whether EKU is a signer option or a signature option.
+    # TODO: Figure out whether this sign_alg is correct for capsules.
+    signature_options = {
+        'sign_alg': 'pkcs12',
+        'hash_alg': 'sha256'
+    }
+    # Set or override OID.
+    signer_options['oid'] = '1.2.840.113549.1.7.2'
```
spbrogan: would like to see this OID defined rather that number text used as it requires look up.

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,162 @@
...
+
+def get_default_arch():
+    return 'amd64'
+
+
+def get_default_os_string():
+    return 'Win10'
+
+
+def build_capsule(capsule_data, capsule_options, signer_module, signer_options):
```
spbrogan: i see no way to add payloads or add drivers.  Does this tool assume only 1 payload and 0 drivers.  I am ok with that but i think we need to document that clearly.

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,162 @@
...
+    while (version_string.count('.') < 3):
+        version_string += '.0'
+    return version_string
+
+
+def get_default_arch():
+    return 'amd64'
+
+
+def get_default_os_string():
```
spbrogan: docstrings would be good.  We should land on consistent format.

edk2toolext/capsule/capsule_tool.py
```diff
@@ -0,0 +1,165 @@
...
+
+    parser.add_argument('-o', dest='options_file', type=argparse.FileType('r'),
+                        help='a filesystem path to a json/yaml file to load with default options. will be overriden by any options parameters') # noqa
+    parser.add_argument('-f', dest='save_final_options', default=False, action='store_true',
+                        help='optional flag to request that final tool options be saved in a file in the output directory')                     # noqa
+
+    parser.add_argument('capsule_payload', type=argparse.FileType('rb'),
+                        help='a filesystem path to the binary payload for the capsule')
+    parser.add_argument('output_dir',
+                        help='a filesystem path to the directory to save output files. if directory does not exist, entire directory path will be created') # noqa
```
spbrogan: what happens if exists and collision?

requirements.txt
spbrogan: https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies
Example of usage. 
https://github.com/psf/requests/blob/3c1d36b827417fdeaf5a1c106129de30dac371d7/setup.py#L98

spbrogan:
because you put your tests in your own folder and not the test folder the tests show up in coverage results. 

For this project we have a test folder at the root.  You can create subfolders in there but you should follow the pattern of the project.  
https://github.com/tianocore/edk2-pytool-extensions/tree/master/edk2toolext/tests

spbrogan:
please look at your code coverage and see if there are practical ways to improve it. https://dev.azure.com/tianocore/edk2-pytool-extensions/_build/results?buildId=1759&_a=summary&tab=ms.vss-codecoverage-web.code-coverage-details&view=codecoverage-tab

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,162 @@
...
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+# OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+# ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+##
+
+
+import uuid
+import os
+# TODO: Replace this step with a new function from within
```
matthewfcarlson: Does this TODO need to be resolved?

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,162 @@
...
+from edk2toollib.uefi.edk2.fmp_payload_header import FmpPayloadHeaderClass
+
+
+def get_capsule_file_name(capsule_options):
+    return f"{capsule_options['fw_name']}_{capsule_options['fw_version_string']}.bin"
+
+
+def get_normalized_version_string(version_string):
+    # 19H1 HLK requires a 4 digit version string, or it will fail
+    while (version_string.count('.') < 3):
```
matthewfcarlson: should we check for non numerical characters?

edk2toolext/capsule/capsule_helper.py
corthon: Eventually, but it really only gets resolved with more changes in edk2pytoollib.
I'll switch these to issues against the two project.

edk2toolext/capsule/capsule_helper.py
corthon: I think I'll leave that be for now. The underlying tools will be validating for and informing the caller of any problem there, but they wouldn't expand the version this way (which has caused problems in the past).

corthon:
@spbrogan, Bueller?

edk2toolext/capsule/capsule_helper.py
```diff
@@ -0,0 +1,185 @@
...
+# https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.pkcs.contentinfo.-ctor?view=netframework-4.8
+PKCS7_SIGNED_DATA_OID = '1.2.840.113549.1.7.2'
+
+
+def get_capsule_file_name(capsule_options):
+    '''from the shared capsule_options dictionary, returns the formatted capsule file name'''
+    return f"{capsule_options['fw_name']}_{capsule_options['fw_version_string']}.bin"
+
+
+def get_normalized_version_string(version_string):
```
spbrogan: not critical...but seems like this really belongs in cat gen tool rather than here since this code should be somewhat agnostic from windows capsules

requirements.txt
spbrogan: does pyopenssl really need to be here?  seems like the extras_require list is the more appropriate way

spbrogan:
i removed my block on the PR.  I did put in a few comments that you should look at.

requirements.txt
corthon: It's in the extras_require, but extras_require is only used when installing modules from other modules (apparently). This is here so that the CI unit testing can reference these modules and get the code coverage up.

edk2toolext/capsule/capsule_helper.py
corthon: I tend to agree. To get the code to land (since the InfGenerator is in library, not extensions), I'll keep it here for now, but will open an issue to move it.

PR #102, opened 2019-10-31
Added path to version aggregator to no longer emit warnings as often 

### LINKED ISSUE #101: Version Aggregator is constantly warning about duplicate entries

matthewfcarlson:
fixes #101


corthon:
The root cause of this is the double update?

matthewfcarlson:
There's definitely some other issue at play here. We don't do a double init of the SDE in platform builds. I think this is the right move regardless because this solves some other problems, but I might want to investigate this a little bit more

edk2toolext/environment/version_aggregator.py
```diff
@@ -29,20 +29,22 @@ def ReportVersion(self, key, value, versionType):
...
-                self._logger.warning("version_aggregator: This {0}:{1} key/value pair "
-                                     "was already registered".format(key, value))
+            old_version = self.Versions[key]
+            if old_version["version"] == value and old_version["path"] == path:
+                pass
             else:
                 error = "version_aggregator: {0} key registered with a different value\n\t" \
-                        "Old:{1}\n\tNew:{2}".format(key, self.Versions[key]["version"], value)
+                        "Old:{1}@{3}\n\tNew:{2}@{4}\n".format(
+                            key, old_version["version"], value, old_version["path"], path)
```
corthon: Is this fine to print if "path" is left as None? Was the None case verified?

matthewfcarlson: Confirmed that None case works fine

matthewfcarlson: ValueError: Intel.ICL.GOP key registered with a different value
        Old:14.0.1027.1@None
        New:14.0.102.1@None

matthewfcarlson:
@corthon None case has been verified

matthewfcarlson:
Fix the verify function to no longer report to version to aggregator

matthewfcarlson:
@spbrogan fixed

PR #105, opened 2019-11-01
Fix a small error in the NugetDependency test.

corthon:
I think that an update in PyTest made this sensitive in a way it didn't used to be.

### DIFF
--- a/edk2toolext/tests/test_nuget_dependency.py
+++ b/edk2toolext/tests/test_nuget_dependency.py
@@ -110,7 +110,7 @@ class TestNugetDependency(unittest.TestCase):
 
         ext_dep_descriptor = EDF.ExternDepDescriptor(ext_dep_file_path).descriptor_contents
         ext_dep = NugetDependency(ext_dep_descriptor)
-        with self.assertRaises(RuntimeError):
+        with self.assertRaises(ValueError):
             # we should throw an exception because we don't know how to parse the version
             ext_dep.fetch()
         self.assertFalse(ext_dep.verify())


corthon:
Okay, this has gotten weird. The test fails locally, but passes on the server. Then I make this commit and it passes locally and fails on the server. Will continue to investigate.

edk2toolext/tests/test_nuget_dependency.py
```diff
@@ -110,7 +110,7 @@ def test_download_bad_nuget(self):
 
         ext_dep_descriptor = EDF.ExternDepDescriptor(ext_dep_file_path).descriptor_contents
         ext_dep = NugetDependency(ext_dep_descriptor)
-        with self.assertRaises(RuntimeError):
+        with self.assertRaises(ValueError):
```
matthewfcarlson: I don't see why this would raise a ValueError?

matthewfcarlson: We only normalize the version when we fetch from cache

matthewfcarlson: My guess is here: https://github.com/tianocore/edk2-pytool-extensions/blob/4d881f5f4bf13baffa444eabc258cf5723e690da/edk2toolext/environment/extdeptypes/nuget_dependency.py#L95

matthewfcarlson: If we don't find a global cache, we bail. My guess is the server doesn't have one. So it bails and blows up with a RuntimeError later in fetch. Your local build has a cache and blows up earlier when it tries to normalize the version. I think we could either normalize the version earlier, or catch either exception.

corthon:
Closing this and opening it as an issue.

PR #108, opened 2019-11-03
Display location of log file after platform build

### LINKED ISSUE #69: location of build log is not displayed on error conditions in build

matthewfcarlson:
Fixes #69

### DIFF
--- a/edk2toolext/base_abstract_invocable.py
+++ b/edk2toolext/base_abstract_invocable.py
@@ -84,7 +84,9 @@ class BaseAbstractInvocable(object):
         log_directory = os.path.join(self.GetWorkspaceRoot(), self.GetLoggingFolderRelativeToRoot())
 
         txtlogfile = self.GetLoggingLevel("txt")
+        self.log_filename = None
         if(txtlogfile is not None):
+            self.log_filename = os.path.realpath(os.path.join(log_directory, self.GetLoggingFileName("txt")))
             logfile, filelogger = edk2_logging.setup_txt_logger(log_directory,
                                                                 self.GetLoggingFileName("txt"),
                                                                 txtlogfile)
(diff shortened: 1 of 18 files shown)


edk2toolext/base_abstract_invocable.py
```diff
@@ -84,10 +84,12 @@ def ConfigureLogging(self):
         log_directory = os.path.join(self.GetWorkspaceRoot(), self.GetLoggingFolderRelativeToRoot())
 
         txtlogfile = self.GetLoggingLevel("txt")
+        self.log_filename = None
```
corthon: Since this field isn't created in the initializer, and nothing tests for None, why initialize it here?

matthewfcarlson: Hmmm- I'll move it to the constructor. It's just a little weird since it is only used by the platform build

matthewfcarlson: I'm open to other ideas of how to solve this problem

corthon: Sorry, the point of my question was: why initialize at all? Can't you just set it if txtlogfile is not None? Does it cause some other problem to not have this field? If ConfigureLogging never happens, there's no way it could be set anyway. Might as well just make the rest of the code tolerant of it not being set. *Or*, if it is a first-class element that could be referenced outside this class, initialize it in the constructor so that it's ubiquitous.

matthewfcarlson: I see what you're saying. I think I might lean towards catching the case where it doesn't exist. I don't know if I want it to be a first class member.

matthewfcarlson: Actually- I retract that. Let's make it a first class member.

matthewfcarlson: Let me know what you think

PR #113, opened 2019-11-05
Update CI Tests for edk2 invokables

matthewfcarlson:


edk2toolext/tests/test_edk2_update.py
```diff
@@ -0,0 +1,117 @@
...
+    def GetPackagesSupported(self):
+        return []
+
+    def GetTargetsSupported(self):
+        return []
+'''
+
+# TODO: move to something lighter weight that has external dependencies
+# Perhaps create a custom nuget package that has itself in it?
+basecore_ext_dep = '''
```
spbrogan: i think you should move to the octocat sample repo or if you need ext_deps in it we can make a test repo based off octocat.  Mu Basecore has got to waste a lot of time and make the unit test not so light weight.

edk2toolext/tests/test_edk2_update.py
```diff
@@ -0,0 +1,117 @@
...
+        extdep_filepath = os.path.abspath(os.path.join(WORKSPACE, "basecore_ext_dep.yaml"))
+        real_class = class_info.replace("WORKSPACE", f'"{workspace_abs}"')
+
+        # inject the settings file and the ext dep into our temp workspace
+        TestEdk2Update.write_to_file(extdep_filepath, basecore_ext_dep)
+        TestEdk2Update.write_to_file(settings_filepath, real_class)
+        sys.argv = ["stuart_update", "-c", settings_filepath]
+        try:
+            builder.Invoke()
+        except SystemExit as e:
```
spbrogan: i don't understand the except case in this unit test

matthewfcarlson: Because we'll always do a system exit, which is an exception in python. So we need to catch it.

edk2toolext/tests/test_edk2_update.py
matthewfcarlson: Agreed- either an octocat sample repo or a nuget package.

edk2toolext/tests/test_edk2_update.py
spbrogan: so this is actually keeps the objects in process and can then continue the test.  Without this the test would just stop here?  Or does pytest/unittest catch this?

matthewfcarlson: According to some documentation, it makes it seem like pytest will catch it, but it will fail the test. I was worried that it would stop all the testing since SystemExit doesn't inherit from Exception (they both come from BaseException)

edk2toolext/tests/test_edk2_update.py
```diff
@@ -48,16 +74,102 @@ def updateClass(cls):
...
+        }}
+        '''
+        # now add the ext_dep to the git repo
+        TestEdk2Update.write_to_file(os.path.join(WORKSPACE, "myrepo_ext_dep.json"), local_ext_dep)
+
+        # Do the update
+        self.invoke_update(settings_filepath)
+
+        # now we should check that our workspace is how we expect it
+        (build_env, shell_env) = self_describing_environment.BootstrapEnvironment(WORKSPACE, [])
```
matthewfcarlson: This should be using the Edk2Update not the SDE directly

corthon:
What's up with this one?

corthon:
@matthewfcarlson ?

matthewfcarlson:
Sorry Bret, I forgot about this. I intend to get to it once I get some pytools time

matthewfcarlson:
@spbrogan and @corthon this should be good to go. It takes about 3-6 seconds on my laptop, depending on how things are cached. I used some of the nuget packages that are used for other tests, so hopefully we'll get some cache benefits

spbrogan:
Matt i hate to do this...but the changes look like they fix numerous bugs. This is a case where the PR has too many things to be squashed as one.  If we are going to encourage proper use of squash merge with the ecosystem we need to walk the walk.    Please determine the right way to partition all the changes and submit the PRs.  I think we can close those pretty quickly.

spbrogan:
Another comment.  I am not a big fan of carrying the content in this https://github.com/matthewfcarlson/edk2-pytool-extensions/tree/test/edk2-update/edk2toolext/tests/minimal_uefi_tree

I prefer the pattern where the test file has that content in a string and writes it to a file.  Most of those look pretty trivial.  Is there a reason you went with this pattern?

matthewfcarlson:
So we're in a bit of a pickle here Sean. I can't check it in because of flake issues. If I fix the flake issue, I need to add code coverage on the things I fixed (you need to match the coverage level on the repo in your diff). I'll separate out the coverage and flake stuff

matthewfcarlson:
Sean- I removed the minimal tree and added a new helper class uefi_tree to the tests. It basically creates a minimal uefi tree. It makes sense to put it all in once place since there is an unhealthy amount of copy and paste inbetween all the test.

edk2toolext/environment/version_aggregator.py
```diff
@@ -17,7 +17,7 @@
 class version_aggregator(object):
```
spbrogan, line 17: what was the reason for this file changes?  Just better structure?  or was their a functional change?

matthewfcarlson: I think better structure- making it a little more clear that you're not supposed to touch the Versions variable

spbrogan: sounds good.  I think it would be good in future to structure your commit to indicate that.  Helps me as the reviewer and then also lets us decide if ok in this PR.  I value our ability to refactor as we go but we should be clear and at times we all run a little fast/loose. :) 

Overall i am good.  Feel free to complete the PR with a good commit message.

matthewfcarlson: No I agree- a better explanation in the commit would have been helpful

matthewfcarlson:
Looks like repo resolver failed :/

spbrogan:
You didn't make changes here, right?  

I have notices github is having service issues.  My guess is it failed to clone due to that

matthewfcarlson:
Nope- retrying

Issue #120, opened 2019-11-21
FDF parser does not import build environment from DSCs


out0xb2:
We fail to parse FDFs that depend on DEFINES defined in the corresponding DSC

**To Reproduce**
I'll post HOWTO later

**Expected behavior**
I can build the KabyLakeRvp, but it breaks

**Execution environment**
List the OS, python version, system environment variables, etc.

**Pip Packages**
If python related run `python -m pip list` and copy the output here

**Additional context**
DEBUG - Parse Active Platform DSC file: C:\git\edk2-platforms-pytools\Platform\Intel\KabylakeOpenBoardPkg/KabylakeRvp3/OpenBoardPkg.dsc
DEBUG - Getting all build keys for build type DEBUG
DEBUG - Parse Active Flash Definition (FDF) file
DEBUG - Getting all build keys for build type DEBUG
ERROR - Invalid file path C:\git\edk2-platforms-pytools\$(PLATFORM_PACKAGE)/Include/Fdf/CorePreMemoryInclude.fdf
CRITICAL - Build Process Exception
ERROR - Traceback (most recent call last):
  File "c:\git\venv_edk2-pytools\lib\site-packages\edk2toolext\environment\uefi_build.py", line 109, in Go
    ret = self.SetEnv()
  File "c:\git\venv_edk2-pytools\lib\site-packages\edk2toolext\environment\uefi_build.py", line 384, in SetEnv
    ret = self.ParseFdfFile()
  File "c:\git\venv_edk2-pytools\lib\site-packages\edk2toolext\environment\uefi_build.py", line 519, in ParseFdfFile
    fdfp.ParseFile(pa)
  File "c:\git\venv_edk2-pytools\lib\site-packages\edk2toollib\uefi\edk2\parsers\dsc_parser.py", line 262, in ParseFile
    self.__ProcessDefines(file_lines)
  File "c:\git\venv_edk2-pytools\lib\site-packages\edk2toollib\uefi\edk2\parsers\dsc_parser.py", line 246, in __ProcessDefines
    (line, add) = self.__ParseDefineLine(l)
  File "c:\git\venv_edk2-pytools\lib\site-packages\edk2toollib\uefi\edk2\parsers\dsc_parser.py", line 184, in __ParseDefineLine
    lf = open(sp, "r")
[...]

PR #121, opened 2019-11-21
UefiBuild must share the environment settings between parsing Dsc and Fdf.  DSC may define macros needed in FDF parsing

out0xb2:
Fixes issue #120 .


out0xb2:
@spbrogan & @matthewfcarlson , take a look!

out0xb2:
@spbrogan 

> I would like to see more comments for this change. I think a description of self.DscEnv would be helpful as well a comment in ParseFdfFile.
> 
> it also would be important to note why the GetAllBuildKeyValues

The content was yours... send me a PR?

matthewfcarlson:
@out0xb2 @spbrogan I'm going to tag and release a new pytool tonight. Since it looks like this is really close, I'll hold off until this is done

PR #136, opened 2019-12-13
[Bug] added super calls for argument parsing

matthewfcarlson:
When doing argument calls, call the next class in the MRO in case we're doing a UEFI build or you have another module.

### DIFF
--- a/edk2toolext/invocables/edk2_setup.py
+++ b/edk2toolext/invocables/edk2_setup.py
@@ -50,11 +50,17 @@ class SetupSettingsManager(MultiPkgAwareSettingsInterface):
 
     def AddCommandLineOptions(self, parserObj):
         ''' Implement in subclass to add command line options to the argparser '''
-        pass
+        try:
+            super().AddCommandLineOptions(parserObj)
+        except AttributeError:
+            pass
 
     def RetrieveCommandLineOptions(self, args):
         '''  Implement in subclass to retrieve command line options from the argparser '''
-        pass
+        try:  # try to call super class just in case there's another module in the MRO
+            super().RetrieveCommandLineOptions(args)
+        except AttributeError:
+            pass
 
     def GetLoggingLevel(self, loggerType):
         ''' Get the logging level for a given type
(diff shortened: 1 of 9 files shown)


edk2toolext/invocables/edk2_setup.py
```diff
@@ -75,7 +81,7 @@ def AddCommandLineOptions(self, parserObj):
         parserObj.add_argument('--omnicache', '--OMNICACHE', '--Omnicache', dest='omnicache_path',
                                default=os.environ.get('OMNICACHE_PATH'))
 
-        super().AddCommandLineOptions(parserObj)
```
corthon: Unneeded newlines.

matthewfcarlson: Fixed

spbrogan:
I really don't like this.  

Two things...I don't think it makes sense.  Can you point me to the python documentation that says that each super class needs to call super otherwise the subclass call to super will not get to all of its super classes.  

two.  I had all the try except blocks that ignore the error.  These objects should not be calling arbitrary things with "super" and thus we should know if the super is going to raise exception or not.

spbrogan:
can read about it here in this 3 part article. 
http://sixty-north.com/blog/the-super-mystery-resolved.html

matthewfcarlson:
I don't think this PR fixes the issue. It just causes everything to add the parameters it wants rather than missing out on certain settings manager classes.

matthewfcarlson:
So here's the fundamental issue we want to solve.

When calling AddCommandLineOptions, we want to call the PlatformBuilder/SettingsClass itself, the relevant settings class, and it's respective settings class parents. Right now, we either don't call the class at all or we don't call the right one during the right phase. This works okay in the case that we have separate PlatformBuilder class from the SettingsClass as we explicitly call out to the UefiBuilder. When they're all combined together is when the problem becomes visible.

For example, for:
```python
 class REDACTEDSettingsManager(UpdateSettingsManager, SetupSettingsManager, BuildSettingsManager, UefiBuilder):
        ''' Add command line options to the argparser '''
        super().AddCommandLineOptions(parserObj)
```
We call UpdateSettingsManager regardless of what invocable we're doing.

matthewfcarlson:
We'd need to swap the desired class once we find it

PR #137, opened 2019-12-14
[Feature] Alternative way to do super calls for commandline parsing

matthewfcarlson:


matthewfcarlson:
The settings class would look like this:
```python
        ''' Add command line options to the argparser '''
        self.SettingsClass.AddCommandLineOptions(self, parserObj)
        if self.SettingsClass == BuildSettingsManager:
            UefiBuilder.AddCommandLineOptions(self, parserObj)
        pass

    def RetrieveCommandLineOptions(self, args):
        '''  Retrieve command line options from the argparser '''
        self.SettingsClass.RetrieveCommandLineOptions(self, args)
        if self.SettingsClass == BuildSettingsManager:
            UefiBuilder.RetrieveCommandLineOptions(self, args)
        pass
```

edk2toolext/edk2_invocable.py
```diff
@@ -126,6 +126,7 @@ def ParseCommandLineOptions(self):
             self.PlatformModule = import_module_by_file_name(os.path.abspath(settingsArg.platform_module))
             self.PlatformSettings = locate_class_in_module(
                 self.PlatformModule, self.GetSettingsClass())()
+            self.PlatformSettings.SettingsClass = self.GetSettingsClass()
```
matthewfcarlson: This needs to get documented

corthon: If we already have a method for this, why are we caching it in a variable?

edk2toolext/invocables/edk2_multipkg_aware_invocable.py
```diff
@@ -20,6 +20,19 @@ class MultiPkgAwareSettingsInterface():
         and which functions can be implemented.
      '''
 
+    SettingsClass = None  # this will be filled in by the edk invocable
```
corthon: Why cache if we already have a method? Is it important to work around the MRO?

matthewfcarlson: No, I just put it in there so it will be populated by autocomplete and it shows up as a class member

edk2toolext/edk2_invocable.py
matthewfcarlson: So PlatformSettings doesn't have a way to get the current Settings Class. This is giving the settings the class that we looked for. The invocable itself has the method.

matthewfcarlson: Documented

edk2toolext/invocables/edk2_multipkg_aware_invocable.py
matthewfcarlson: Since we don't have access to the GetSettingsClass method in the SettingsManager, this is here to cache it inside the settings

corthon: Still need this after the new change?

matthewfcarlson: Nope. Good catch

PR #141, opened 2019-12-20
Pass same build variables to DSC parser as to FDF parser.

### LINKED ISSUE #140: DSC parser should be passed input values from previous parsing steps.
DSC Parser Should be instantiated with values from prior in the parsing project (in particular, from target.txt). Currently it cannot resolve e.g. $(TOOL_CHAIN_TAG). 

**To Reproduce**

Introduce a conditional into the build like the following
!if ("GCC49" in $(TOOL_CHAIN_TAG)) OR ("GCC5" in $(TOOL_CHAIN_TAG))
asdfsadf
!endif

**Expected behavior**
Conditional is properly evaluated. But instead, stack trace.

joschock:
Fixes #140

### DIFF
--- a/edk2toolext/environment/uefi_build.py
+++ b/edk2toolext/environment/uefi_build.py
@@ -486,8 +486,14 @@ class UefiBuilder(object):
             # parse DSC File
             logging.debug(
                 "Parse Active Platform DSC file: {0}".format(dsc_file_path))
+
+            # Get the vars from the environment that are not build keys
+            input_vars = self.env.GetAllNonBuildKeyValues()
+            # Update with special environment set build keys
+            input_vars.update(self.env.GetAllBuildKeyValues())
+
             dscp = DscParser().SetBaseAbsPath(self.ws).SetPackagePaths(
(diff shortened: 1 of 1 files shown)


matthewfcarlson:
I think this is consistent with #121

matthewfcarlson:
@joschock how have you tested this?

joschock:
Yep - fixes my issue with the problematic DSC conditional. Also rebuilt several projects with this tool, observed no issues.

matthewfcarlson:
I think there's some more work that needs to be done here. But it's approved and you can merge when you're ready

joschock:
Agree; but I think most of that work is deciding/designing how the parser works and what it is allowed to see. I think that should be done as a larger separate effort.

joschock:
I don't have rights to complete the PR.

matthewfcarlson:
Sorry, when I said there was more working to be done I meant more design needed to be done. Our whole approach to passing information from the DSC to the fdf needs to be thought through a little closer

PR #155, opened 2020-02-20
[Feature] added local version feature for ext_deps

matthewfcarlson:
This allows a developer to force the SDE to accept a local version.


edk2toolext/environment/version_aggregator.py
```diff
@@ -15,6 +15,9 @@
 
 
 class version_aggregator(object):
+
+    LOCAL_VERSION = "LOCAL"
```
spbrogan: i don't think this belongs here.  LOCAL is a feature of external_dependency and not version aggregator.

corthon: Agreed. Perhaps instead of Version == LOCAL, it could be Type == LOCAL? (Or Source?)

corthon:
Any thoughts on this one? If it's not ready or we need to pivot, should we close it and come back later?

matthewfcarlson:
There was discussion on whether this was a feature we event wanted to support

spbrogan:
Lets close this for now.  Feel free to open again if the use case gains priority

PR #183, opened 2020-05-01
Feature/code coverage

matthewfcarlson:


spbrogan:
- [ ] Need documentation describing this and how to manage any tokens

- [ ]  Needs to work on both Linux and Windows

- [ ]  Since this template file is used on multiple projects we need to make sure it will not break them.  Maybe it is conditional on the variable existing? 

- [ ] Does codecov.io handle PR (incoming noise) vs CI (baseline)

- [ ] Where is the policy about what is a PR failure?

matthewfcarlson:
@spbrogan I think is ready

matthewfcarlson:
https://codecov.io/gh/tianocore/edk2-pytool-extensions/pull/183

matthewfcarlson:
^ Not as useful since it hasn't run against master, but it would look like that

matthewfcarlson:
@spbrogan or @corthon thoughts? I'd like to get this in and see if we like it

spbrogan:
@matthewfcarlson  - i see you checked the box for Windows / Linux yet it looks like it only runs on Linux?  Is that accurate?

matthewfcarlson:
It will run on both, I just turned off windows to test the parameter. Fixed.

spbrogan:
Do we need to get @mdkinney to install the app into tianocore org to see if it will then allow anonymous access to the coverage results?   I don't think it is useful if a user has to log in to codecov.io and be a member of tianocore to see the results of their PR.

matthewfcarlson:
Yeah- I sent a request to the tianocore org to allow codecov access to this specific repo (you can choose which repos it has access to).

spbrogan:
thanks.  @mdkinney  that fixed it.

PR #185, opened 2020-05-12
[Docs] Auto generation and publishing of docs

matthewfcarlson:


matthewfcarlson:
@spbrogan I think this is the final form. What's next is to build the deployment. Does the pipeline have permission to push to the gh-pages branch?

azure-pipelines/templates/auto-doc-job.yml
```diff
@@ -0,0 +1,54 @@
+# File build-test-job.yml
```
spbrogan, line 1: comment incorrect

spbrogan, line 1: I think i would combine the two yml files.  There is no reason you need to use templates.  You are noting trying to support numerous environments or toplevel files, right?  We do this in other places because we want to reuse the same steps for linux and windows and OSX(if we did this).  that would never be the case for a release like this.

matthewfcarlson:
After giving this a quick thought, I think we need to check this into pytool library, not extensions but check in the azure-pipeline file into here. Seems like a bit of inversion there

PR #193, opened 2020-05-22
Clean up invokable SettingsManagers and hierarchy

spbrogan:
Make APIs more consistent and don't override functions in Settings Managers.
Raise Exceptions for required functions
Set safe defaults for optional functions

### DIFF
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -31,15 +31,15 @@ class Edk2InvocableSettingsInterface():
         and can be implemented in a settings manager.
     '''
 
-    def GetWorkspaceRoot(self):
+    def GetWorkspaceRoot(self) -> str:
         ''' get absolute path to WorkspaceRoot '''
         raise NotImplementedError()
 
-    def GetPackagesPath(self) -> Iterable:
+    def GetPackagesPath(self) -> Iterable[os.PathLike]:
         ''' Optional API to return an Iterable of paths that should be mapped as edk2 PackagesPath '''
         return []
 
-    def GetActiveScopes(self) -> Tuple(str):
+    def GetActiveScopes(self) -> Tuple[str]:
         ''' Optional API to return Tuple containing scopes that should be active for this process '''
         return ()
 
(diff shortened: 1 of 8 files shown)


spbrogan:
Need to fix tests and understand why failing.  My goal was to refactor without breaking anything.  :)

edk2toolext/edk2_invocable.py
```diff
@@ -22,6 +23,44 @@
...
+        txt  == plain text file logging
+        md   == markdown file logging
+        '''
+        return None
+
+    def AddCommandLineOptions(self, parserObj) -> None:
+        ''' Implement in subclass to add command line options to the argparser '''
+        pass
+
+    def RetrieveCommandLineOptions(self, args):
```
matthewfcarlson, line 59: args would be an object (https://docs.python.org/3/library/typing.html)

edk2toolext/edk2_invocable.py
```diff
@@ -22,6 +23,44 @@
...
+        This is an interface definition only
+        to show which functions are required to be implemented
+        and can be implemented in a settings manager.
+    '''
+
+    def GetWorkspaceRoot(self):
+        ''' get absolute path to WorkspaceRoot '''
+        raise NotImplementedError()
+
+    def GetPackagesPath(self) -> Iterable:
```
matthewfcarlson, line 38: Iterable[os.PathLike] would be more specific but not required

edk2toolext/edk2_invocable.py
```diff
@@ -22,6 +23,44 @@
...
+        ''' Optional API to return an Iterable of paths that should be mapped as edk2 PackagesPath '''
+        return []
+
+    def GetActiveScopes(self) -> Tuple(str):
```
matthewfcarlson, line 42: Tuple[str]

edk2toolext/edk2_invocable.py
matthewfcarlson, line 59: -> None

edk2toolext/edk2_invocable.py
```diff
@@ -46,6 +85,12 @@ def GetWorkspaceRoot(self):
         except AttributeError:
             raise RuntimeError("Can't call this before PlatformSettings has been set up!")
 
+    def GetPackagesPath(self):
+        try:
+            return self.PlatformSettings.GetPackagesPath()
+        except AttributeError:
+            raise RuntimeError("Can't call this before PlatformSettings has been set up!")
+
     def GetActiveScopes(self):
```
matthewfcarlson, line 94: -> Iterable[str]

matthewfcarlson: Do you need a tuple or iterable?

edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -25,68 +26,16 @@
...
-        ''' Get the logging level for a given type
-        base == lowest logging level supported
-        con  == Screen logging
-        txt  == plain text file logging
-        md   == markdown file logging
-        '''
-        pass
-
-    def GetPluginSettings(self):
+    def GetPluginSettings(self) -> Dict():
```
matthewfcarlson, line 33: Dict[str, Any]

edk2toolext/edk2_invocable.py
spbrogan: In Edk2Invocable we are doing tuple concatenation so i need a tuple. I tried to find out what would happen if I added a tuple to an iterable …but couldn't find anything good....so decided to just keep it as a tuple.

PR #194, opened 2020-05-22
[Bug] fixed flake/coverage issues

matthewfcarlson:
Recent flake updates have caused a build break. Also locks flake8 to a version so this doesn't happen again.
Also adds a flag to the pipeline file so that runs are tagged on CodeCov.io - reports were being squashed together and it was hard to tell what was being tested on which platforms.


matthewfcarlson:
So I've included two unit tests for the files I touched because of flake issues. Normally I'd check in those unit tests before these fixes but the flake issues are fun

matthewfcarlson:
Sean let me know if you want the unit tests removed. I'll remove them and you can force complete it

PR #212, opened 2020-07-15
VERSIONINFO Tools

gatys:
Provides tooling to decode and encode VERSIONINFO of PE/PE+ files.


.flake8
```diff
@@ -1,5 +1,6 @@
 [flake8]
 #E266 too many leading '#' for block comment
 #E722 do not use bare 'except'
-ignore = E266,E722
+#W503 conflcits with updated rule W504
```
matthewfcarlson, line 4: I'd break this out into a separate PR if really needed

edk2toolext/tests/versioninfo/versioninfo_tool_test.py
```diff
@@ -0,0 +1,99 @@
...
+    def test_encode_decode(self):
+        cli_params = [self.dummy_json, self.temp_dir]
+        parsed_args = versioninfo_tool.get_cli_options(cli_params)
+        versioninfo_tool.serviceRequest(parsed_args)
+        cd = os.getcwd()
+        os.system('cd %s && nmake rsrc && cd %s' % (self.temp_dir, cd))
+        cli_params = [os.path.join(self.temp_dir, DUMMY_EXE_FILE_NAME) + '.exe', self.temp_dir, '-d']
+        parsed_args = versioninfo_tool.get_cli_options(cli_params)
+        versioninfo_tool.serviceRequest(parsed_args)
+        with open(os.path.join(self.temp_dir, 'VERSIONINFO.json')) as generated_json:
```
matthewfcarlson, line 88: Maybe put 'VERSIONINFO.json' in a variable as well?

edk2toolext/versioninfo/versioninfo_tool.py
```diff
@@ -0,0 +1,67 @@
...
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+
+import os
+import sys
+import argparse
+import json
+from edk2toolext.versioninfo.versioninfo_helper import PEObject, VERSIONINFOGenerator
```
matthewfcarlson, line 16: I don't love the name VersionInfo, thoughts @spbrogan? I feel like it would get mixed up with version aggregator.

edk2toolext/versioninfo/versioninfo_tool.py
```diff
@@ -0,0 +1,67 @@
...
+
+TOOL_DESCRIPTION = """
+Versioninfo Tool is a command-line tool to assist in generating VERSIONINFO
+resource files for use with Resource Compiler. It takes a JSON representing
+versioning info and produces a resource file that satifies UEFI SBAT requirements
+and is compatible with Resource Compiler.
+
+An example call might look like:
+%s -o /path/to/version.JSON /path/to/output
```
matthewfcarlson, line 25: so since this is in extensions, we'll expose a console script. So it'll be binver_encode or binver_decode or something like that

setup.py
```diff
@@ -71,7 +71,8 @@ def run(self):
                             'stuart_pr_eval=edk2toolext.invocables.edk2_pr_eval:main',
                             'omnicache=edk2toolext.omnicache:main',
                             'nuget-publish=edk2toolext.nuget_publishing:go',
-                            'edk2_capsule_tool=edk2toolext.capsule.capsule_tool:main']
+                            'edk2_capsule_tool=edk2toolext.capsule.capsule_tool:main',
+                            'versioninfo_tool=edk2toolext.versioninfo.versioninfo_tool:main']
```
matthewfcarlson, line 75: versioninfo_tool is good

edk2toolext/tests/versioninfo/versioninfo_tool_test.py
```diff
@@ -0,0 +1,99 @@
...
+            generatedDict = json.load(generated_json)
```
matthewfcarlson, line 89: snake case please 🐍

matthewfcarlson, line 89: you also might need to do a json safeload here

setup.py
```diff
@@ -71,7 +71,8 @@ def run(self):
...
     },
     install_requires=[
```
spbrogan, line 77: don't we need pefile listed here?  The pip requirements listed in txt file are for development and publishing.  For actual runtime we need to list dependencies here.

.flake8
spbrogan: this change is fine as it is also in pytool-library.  
I would suggest we align the comment to match.
It looks like this:   
#W503 line break before binary operator, see https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

see https://github.com/tianocore/edk2-pytool-library/blob/master/.flake8

edk2toolext/versioninfo/versioninfo_tool.py
```diff
@@ -0,0 +1,70 @@
...
+    command_group.add_argument('-d', '--dump', action='store_const', dest='mode', const='d',
+                               help='outputs json file of VERSIONINFO given PE file')
+    parser.set_defaults(mode='e')
+    return parser.parse_args(args=args)
+
+
+def serviceRequest(args):
+    output_dir = args.output_dir
+    if not output_dir.endswith('\\'):
+        output_dir += '\\'
```
spbrogan, line 52: this is windows specific.  either use os.sep or figure out better path handling.

edk2toolext/versioninfo/versioninfo_tool.py
```diff
@@ -0,0 +1,70 @@
...
+    if args.mode == 'd':
+        pe = PEObject(args.input_file)
```
spbrogan, line 54: might be helpful to check if file exists so you can provide a nice error code for a common error

edk2toolext/versioninfo/versioninfo_tool.py
```diff
@@ -0,0 +1,70 @@
...
+        output_dir += "VERSIONINFO.json"
```
spbrogan, line 55: os.path.join is better pattern.

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+}
+
+
+class PEObject(object):
+    pe = None
+
+    # Takes filepath to PE/PE+ file to parse
+    def __init__(self, filepath):
+        try:
+            print(filepath)
```
spbrogan, line 240: should use logging instead of print so that it is captured inside all logs rather than just command line console

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+            self.pe = pefile.PE(filepath)
+        except FileNotFoundError:
+            print("ERROR: Could not find " + filepath, file=sys.stderr)
+
+    # Private, given a 32 bit word, converts word into a string of 2 16 bit integers
+    # seperated by '.' to conform to version string format.
+    def __versionStr__(self, val):
```
spbrogan, line 247: can you use python tying syntax

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+            except AttributeError:
+                print("WARNING: Could not find FileInfoTable in .rsrc section.", file=sys.stderr)
+        else:
+            print("WARNING: Could not find .rsrc section.", file=sys.stderr)
+
+        return result
+
+
+# Given a JSON file, object creates generator for VERSIONINFO rc file for the given JSON.
+# Provides validaiton of some basic requirements and generates VERSIONINFO.rc source to be used with resource compiler.
```
spbrogan, line 360: spelling

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+            if field == TRANSLATION_STR:
+                outStr += "\t\t" + VALUE_STR + " \"" + field + "\",\t" + languageTokens[0] + "," \
+                          + languageTokens[1] + "\n"
+            else:
+                outStr += "\t\t" + VALUE_STR + " \"" + field + "\",\t\"" \
+                          + self.versionDict[VAR_FILE_INFO_STR][field] + "\"\n"
+
+        outStr += "\t" + END_STR + "\n" + END_STR + "\n#endif"
+
+        with open(path + "VERSIONINFO.rc", "w") as out:
```
spbrogan, line 555: os.path.join(path, "VERSIONINFO.rc") would be better

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+            except ValueError:
+                print("ERROR: Invalid language code: " + langIDset[0] + "\".", file=sys.stderr)
+                valid = False
+        else:
+            print("ERROR: Missing required parameter: Translation in VarFileInfo", file=sys.stderr)
+            valid = False
+
+        return valid
+
+    def write(self, path):
```
spbrogan, line 507: docstring

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+        if not self.validate():
+            return
```
spbrogan, line 509: logging or error message?

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+
+        outStr = "/* Auto-generated VERSIONINFO resource file. */\n\n"
```
spbrogan, line 511: might be good to add timestamp as well as tool version that way debugging a tool error is easier at a later point in time when there are multiple versions of the tool

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+import pefile
+import sys
+import json
+import copy
+
+# String values for PE/PE+ header versioning metadata
+FILE_OS_STRINGS = {
```
spbrogan, line 18: there is a lot of global data here.  Does it make sense to wrap this inside of a class?

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+            else:
+                if key.upper() in required:
+                    required.remove(key.upper())
+
+        for remaining in required:
+            print("ERROR: Missing required parameter: " + remaining + ".", file=sys.stderr)
+            valid = False
+
+        if STRING_FILE_INFO_STR in self.versionDict:
+            requiredStringFileFields = copy.deepcopy(STRING_FILE_INFO_REQUIRED_FIELDS)
```
spbrogan, line 439: another trick for this that doesn't require importing more code is to use the slice operation.  STRING_FILE_INFO_REQUIRED_FIELDS[:] will create a copy you can delete from.

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+class PEObject(object):
+    pe = None
+
+    # Takes filepath to PE/PE+ file to parse
+    def __init__(self, filepath):
+        try:
+            print(filepath)
+            self.pe = pefile.PE(filepath)
+        except FileNotFoundError:
+            print("ERROR: Could not find " + filepath, file=sys.stderr)
```
spbrogan, line 243: if you use logging you don't have to deal with this. sys.stderr stuff.

edk2toolext/versioninfo/versioninfo_helper.py
```diff
@@ -0,0 +1,556 @@
...
+# PE/PE+ files.
+#
+##
+# Copyright (C) Microsoft Corporation
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+import pefile
+import sys
```
spbrogan, line 13: not a fan of importing sys into a non cli module.  Looks like logging could remove your dependency on sys.

edk2toolext/tests/versioninfo_tool_test.py
```diff
@@ -0,0 +1,217 @@
...
+        with open(dummy_exe_src, 'w') as dummy_src:
+            dummy_src.write(DUMMY_EXE_SOURCE)
+        with open(dummy_exe_makefile, 'w') as dummy_makefile:
+            dummy_makefile.write(DUMMY_EXE_MAKEFILE)
+
+        cli_params = [dummy_json, temp_dir]
+        parsed_args = versioninfo_tool.get_cli_options(cli_params)
+        versioninfo_tool.service_request(parsed_args)
+        cd = os.getcwd()
+        os.system('cd %s && nmake rsrc && cd %s' % (temp_dir, cd))
```
matthewfcarlson, line 143: Probably better to use RunCmd here. You can pass in the directory you want it to execute in

edk2toolext/versioninfo/versioninfo_tool.py
matthewfcarlson: @spbrogan @gatys Any thoughts on this before we merge this in?

spbrogan: it is really binary info, image info, PE info, resource info?? It holds more than version.  Not sure if that would make this better.

spbrogan:
Two additional comments. 
1. I don't see python typing being used.  I think for new code we should be doing this at least for simple types.  
2. I don't see unit tests that cover many of the error/invalid paths.  Looking at coverage i see the good path followed but most else conditions and exception paths are untested.  This is ok for functionality but doesn't lead to a very robust tool.  Often as usage scales out beyond the developer there will be lots of questions and problems because the user doesn't know how to use it exactly right.

PR #213, opened 2020-08-19
fix web dependency unzipping blank folder

matthewfcarlson:


edk2toolext/environment/extdeptypes/web_dependency.py
```diff
@@ -129,9 +129,6 @@ def fetch(self):
         # # If we're unpacking a directory, we can copy the important parts into
         # # a directory named self.contents_dir
         if self.download_is_directory:
-            # The root of the internal path is the folder we will see populated in descriptor_location
-            unzip_root = WebDependency.get_internal_path_root(self.descriptor_location, self.internal_path)
-
             logging.info(f"Copying directory from {complete_internal_path} to {self.contents_dir}")
```
spbrogan, line 132: It looks like your change is removing the optimization that only copies part of the archive.

matthewfcarlson: I don't think that's the case. I think this just deletes more of the temporary folder and in the correct location

matthewfcarlson: Windows Robot passed. Manual EDK2 build under WSL seemed fine as well (I just built one package)

matthewfcarlson: I'll test how long it takes to clone

PR #216, opened 2020-09-16
Added extra information to extdep names when present

matthewfcarlson:
Currently, there isn't a good way to add ext_deps that point to the same resource at different versions. This adds the ID to the name if present to prevent version aggregator from exploding


matthewfcarlson:
@spbrogan or @out0xb2 this is blocking some folks trying to work around this

matthewfcarlson:
Whoops, sorry Jeremiah, I meant @corthon

Issue #228, opened 2020-11-02
NugetDependency._fetch_from_cache() is no longer working


corthon:
https://github.com/tianocore/edk2-pytool-extensions/blob/37d367816349270e4842864cd8883efdadea682e/edk2toolext/environment/extdeptypes/nuget_dependency.py#L126

Suspect that Nuget has changed its cache store and the extra 'package_name' can be dropped from the end.

matthewfcarlson:
Excellent. Good diagnosis.

What's the impact? Is this linux only or windows? Does this affect the usage of the system?

corthon:
I noticed it on Windows. Checking Linux...

corthon:
Hmmm... suggested fix might be incorrect. I've got one cached package called "content" and another with package_name (as the final directory name).

corthon:
It *may* be that we can just assume whatever directory under that version directory is the content.

corthon:
Hmmm... this is an odd assortment. I wonder how they published this one that has an inner "content" directory.

corthon:
Okay, so here's a proposed fix. This won't do bad things if the inner cache doesn't look right, but it will at least let the caller know that there's a cache that isn't being used.

```patch
diff --git a/edk2toolext/environment/extdeptypes/nuget_dependency.py b/edk2toolext/environment/extdeptypes/nuget_dependency.py
index dd34d0c..3c797ea 100644
--- a/edk2toolext/environment/extdeptypes/nuget_dependency.py
+++ b/edk2toolext/environment/extdeptypes/nuget_dependency.py
@@ -122,14 +122,22 @@ def _fetch_from_cache(self, package_name):
         #
         # Now, try to locate our actual cache path
         nuget_version = NugetDependency.normalize_version(self.version)
+
         cache_search_path = os.path.join(
-            self.global_cache_path, package_name.lower(), nuget_version, package_name)
+            self.global_cache_path, package_name.lower(), nuget_version)
+        inner_cache_search_path = os.path.join(cache_search_path, package_name)
         if os.path.isdir(cache_search_path):
-            logging.info(
-                "Local Cache found for Nuget package '%s'. Skipping fetch.", package_name)
-            shutil.copytree(cache_search_path, self.contents_dir)
-            self.update_state_file()
-            result = True
+            # If we found a cache for this version, let's use it.
+            if os.path.isdir(inner_cache_search_path):
+                logging.info(
[...]

PR #234, opened 2021-01-22
nuget_dependency: Warn on malformed cache

corthon:
If the Nuget Dependency code finds a cache but cannot
heuristically match the package contents (so that they may
be copied directly) produce a warning to the user that they
may be using a malformed package..

Issue #228


corthon:
Details about approach are in the issue discussion. (#228 )

matthewfcarlson (approved):

matthewfcarlson:
I would love a test that creates a malformed cache and make sure we warn the user

PR #242, opened 2021-03-29
Pytools: Add the customization invocable and builder

BobCF:
The CustomizableBuild class is similar to Edk2PlatformBuild, but it consumes CustomizableBuilder.

The CustomizableBuilder class is the base class, its child class do the real implementation.

Signed-off-by: Bob Feng <bob.c.feng@intel.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,8 @@ setuptools.setup(
                             'sig_db_tool=edk2toolext.uefi.sig_db_tool:main',
                             'firmware_policy_tool=edk2toolext.windows.policy.firmware_policy_tool:main',
                             'edk2_capsule_tool=edk2toolext.capsule.capsule_tool:main',
-                            'versioninfo_tool=edk2toolext.versioninfo.versioninfo_tool:main']
+                            'versioninfo_tool=edk2toolext.versioninfo.versioninfo_tool:main',
+                            'stuart_customization_build=edk2toolext.invocables.customizable_build:main']
     },
     install_requires=[
         'pyyaml>=5.3.1',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -74,7 +74,8 @@ def run(self):
                             'sig_db_tool=edk2toolext.uefi.sig_db_tool:main',
                             'firmware_policy_tool=edk2toolext.windows.policy.firmware_policy_tool:main',
                             'edk2_capsule_tool=edk2toolext.capsule.capsule_tool:main',
-                            'versioninfo_tool=edk2toolext.versioninfo.versioninfo_tool:main']
+                            'versioninfo_tool=edk2toolext.versioninfo.versioninfo_tool:main',
+                            'stuart_customization_build=edk2toolext.invocables.customizable_build:main']
```
corthon, line 78: There's a bit of a name change here, which could be confusing.

corthon (commented):
In general, this looks good, but I really need to see how it would be used to understand the usefulness.

setup.py
BobCF: I'll figure out a more proper name.

In the Intel Platform Build process, there are many build steps including 3 times evoke edk2 build instance, and every edk2 build instance has many its own prebuild steps and postbuild steps.  

So I'd need a new stuart builder which can execute customized command rather than only the edk2 build.py.(UefiBuilder is just used to call edk2 build). The new build can be used to implement the prebuild/postbuild steps. And also can be used to implement a higher-level build process scheduler to evoke multiple stuart_build commands.

Javagedes:
Closed due to inactivity, feel free to re-open!

PR #244, opened 2021-03-31
Harden package content move in nuget_dependency fetch

joschock:
This patch adds robustness to the movement of nuget extdep to its final location. Previous code would occasionally fail and raise issues when the source_dir was not completely removed as part of the call to move. This change splits the copy and the remove, and adds retries to the remove operation.


corthon (approved):

edk2toolext/environment/extdeptypes/nuget_dependency.py
```diff
@@ -215,7 +215,17 @@ def fetch(self):
...
+        shutil.copytree(source_dir, self.contents_dir)
+
+        for _ in range(3):  # rmtree doesn't always remove all files. Attempt delete up to 3 times.
+            try:
+                shutil.rmtree(source_dir)
+            except OSError as err:
+                logging.warning(f"Failed to fully remove {source_dir}: {err}")
+            else:
+                break
+        else:
```
corthon, line 227: So Pythonic!

PR #246, opened 2021-04-21
Add parallelization and retries to Omnicache fetch

joschock:
This updates Omnicache fetch to parallelize the fetch of remote repos using -j and --multiple. It also adds retries for failed fetches. This is expected to improve speed and resilience of omnicache.


edk2toolext/omnicache.py
```diff
@@ -403,12 +410,20 @@ def main():
         # as an optimization, if input config file provided, only fetch remotes specified in input config
         # otherwise, fetch all remotes in the OmniCache
         if (input_config_remotes is not None):
-            remotes = (x["name"] for x in input_config_remotes)
```
corthon, line 406: Why did this need to change to a list?

joschock: mostly for consistency with line 415. line 415 had to change it was assigning "remotes" to the keys() view - and that meant that constructing two lists (lines 418/419) was hard - because the first list would exhaust the view iterating over it, leaving the second list empty no matter whether there were items in remote that matched the condition.  I think I could leave this one a tuple, if it matters. I could also change 415 to 
`remotes = omnicace_config.remotes.list()`, I think. But I didn't try that.

joschock: er, that should be `remotes = list(omnicache_config.remotes)` I guess.

corthon: Not that big of a deal. Just checking if there was a specific technical reason. "Consistency" works for me.

corthon (approved):

PR #248, opened 2021-04-22
Resolve LGTM reported errors and warnings

spbrogan:


edk2toolext/tests/test_web_dependency.py
```diff
@@ -211,24 +211,6 @@ def test_sha256_lowercase_single_file(self):
         if not os.path.isfile(file_path):
             self.fail("The downloaded file isn't there")
 
-    # Test that get_internal_path_root works the way we expect with a flat directory structure.
```
corthon, line 214: Why drop the tests?

edk2toolext/environment/uefi_build.py
```diff
@@ -172,10 +172,20 @@ def Go(self, WorkSpace, PackagesPath, PInHelper, PInManager):
                     logging.critical("Flash Image failed")
                     return ret
 
-        except:
+        except Exception:
             logging.critical("Build Process Exception")
             logging.error(traceback.format_exc())
             return -1
+
+        except SystemExit:
```
corthon, line 180: These were suggested additions? Or you decided to expand the checks since you were already hitting this file?

spbrogan: as part of changing from handling all exceptions to handling the three base types (which was the suggestion) I added unique messages so it could be determined what happened.  So it isn't really much of an expansion.

edk2toolext/tests/test_web_dependency.py
spbrogan: The test was dropped because the function was removed.  The function was removed because no one called it in the code other than then test.

corthon (approved):

PR #252, opened 2021-05-05
Refactor OmniCache

joschock:
Refactor Omnicache to improve performance, handle tags better, and remove internal yaml config file.

Omnicache version 0.11 design notes:
1. Only a single remote per URL will be permitted.
2. Remote names inside omnicache will be UUIDs generated when a new URL is added.
3. Users may specify names for remotes, but they will be treated simply as "display names" without git significance.
4. As many objects as possible will be cached for any given remote; so all remotes will fetch tags.
 4a. To avoid tag name collisions between remotes, tags will be fetched into per-remote namespaces.
5. Older omnicaches will be updated to meet the above design points when first encountered by the script.
 5a. This implementation takes care that the underlying git construction is compatible (but not performant) when an
     older omnicache interacts with it. Older omnicache should not crash when encountering a newer omnicache directory
     but may take a lot of slow and unecessary actions to "re-initialize" it.

### DIFF
--- a/edk2toolext/omnicache.py
+++ b/edk2toolext/omnicache.py
@@ -82,6 +82,9 @@ class Omnicache():
             if (out.getvalue().strip() == OMNICACHE_VERSION):
                 logging.debug("{0} - matching omnicache version. valid (convertible don't care).".format(self.path))
                 return (True, True)
+            else:
+                logging.debug("{0} - non-matching omnicache version. not valid (is convertible).".format(self.path))
+                return (False, True)
         logging.debug("{0} - not a bare repo. not valid (not convertible).".format(self.path))
         return (False, False)
 
--- a/edk2toolext/omnicache.py
+++ b/edk2toolext/omnicache.py
@@ -10,250 +10,354 @@ import logging
 import argparse
 import datetime
 import yaml
+import uuid
 from io import StringIO
 
 from edk2toolext import edk2_logging
-from edk2toollib import utility_functions
-from edk2toolext.edk2_git import Repo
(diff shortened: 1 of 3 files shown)


joschock:
rebased to latest.

edk2toolext/omnicache.py
```diff
@@ -10,252 +10,366 @@
...
+#     older omnicache interacts with it. Older omnicache should not crash when encountering a newer omnicache directory
+#     but may take a lot of slow and unecessary actions to "re-initialize" it.
+
+OMNICACHE_VERSION = "0.11"
+
+# Prior to version 0.11, an "omnicache.yaml" file in the root of the omnicache repo was used to manage the cache
+# configuration. Starting with 0.11, git config entries are used directly. If the file below is present, it will be
+# removed to force re-initialization of the omnicache by older omnicache instances. (e.g. if an old branch with a prior
+# implementation of omnicache is being built and uses the same cache directory)
+OMNICACHE_FILENAME = "omnicache.yaml"
```
corthon, line 36: Should we change this CONSTANT name? To maybe PRE_0_11_OMNICACHE_FILENAME or something?

corthon (commented):
Didn't do a detailed review, but finally had some time to take a look at this. I'm good to let it go into 0.14.3, if you think it's ready.

edk2toolext/omnicache.py
joschock: Sure - updated.

joschock:
rebased to latest master.

corthon (approved):

PR #253, opened 2021-06-08
Added logic to support nuget versions suffix with numbers

kuqin12:
As exemplified in nuget versioning site: https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics, the current nuget package supports versioning suffixes such as "alpha2", "beta1", etc.

This change added extra logic to support nuget versions suffix with numbers instead of failing to update with "Unparsable version tag".

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -122,6 +122,7 @@
         "Boothole",
         "dbxupdate",
         "FileFlagsMask",
-        "markdownlint"
+        "markdownlint",
+        "xffd"
     ]
 }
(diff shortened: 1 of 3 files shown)


.cspell.json
```diff
@@ -122,6 +122,7 @@
         "Boothole",
         "dbxupdate",
         "FileFlagsMask",
-        "markdownlint"
+        "markdownlint",
+        "xffd"
```
corthon, line 126: Where was this needed?

corthon (approved):

.cspell.json
kuqin12: This was used in edk2_setup.py. I am not sure why it start to bubble up in this PR given I have not touched it.

PR #254, opened 2021-06-10
capsule: Update CapsuleHelper to process integrity file

corthon:
Update CapsuleHeader to check for the 'fw_integrity_file' capsule option and,
if present, set the field in the InfGenerator so that the INF file has the
definitions for the integrity file.

### DIFF
--- a/edk2toolext/capsule/Readme.md
+++ b/edk2toolext/capsule/Readme.md
@@ -1,0 +1,40 @@
+# Capsule Helper and Tools
+
+These modules are designed to help developers build, sign, and distribute correctly formatted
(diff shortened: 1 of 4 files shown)


spbrogan:
This is going to put a dependency on pytool library version, right?

corthon:
> This is going to put a dependency on pytool library version, right?

Yes. It will. I should hold this and update the requirements once Library lands.

corthon:
This SHOULD be good to go now.

spbrogan (approved):
Approve but seems like adding a new option should have some documentation update somewhere even if it is just a function header.

corthon:
Documentation started. Can fill in more if the tools are EVER consumed.

PR #256, opened 2021-06-14
Add a builder's base class so that the user

BobCF:
can create customized builder and make it be consumed by
edk2PlatformBuild. That is to be consumed by stuart_build.exe

Signed-off-by: Bob Feng <bob.c.feng@intel.com>

### DIFF
--- a/edk2toolext/environment/abs_builder.py
+++ b/edk2toolext/environment/abs_builder.py
@@ -1,0 +1,22 @@
+# @file abs_build.py
+# This module contains code that supports the Tianocore Edk2 build system
+# This class is designed to be subclassed by a platform to allow
+# more extensive and custom behavior.
+##
+# Copyright (c) Microsoft Corporation
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+class AbsBuilder(object):
+
+    # TODO: this initiablize block should go to pipeline builder
+
+    def AddPlatformCommandLineOptions(self, parserObj):
+        pass
+
+    def RetrievePlatformCommandLineOptions(self, args):
+        pass
+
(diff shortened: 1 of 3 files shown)


spbrogan (changes requested):
I don't think this change makes sense.  Basically because your AbsBuilder is completely abstract you are using stuart_build/edk2_platform_build  just to run a python module that has "go".  At that point what value is using stuart_build?

I think I would suggest you create your own invocable of the type want.  I don't see value in this unless you get code reuse or simplification/scalability.  If with each platform or new generation you are implementing a complete AbsBuilder then you haven't achieved those goals.

BobCF:
> I don't think this change makes sense. Basically because your AbsBuilder is completely abstract you are using stuart_build/edk2_platform_build just to run a python module that has "go". At that point what value is using stuart_build?
> 

No, not just run 'go'.  I want to maximize the reuse of the edk2_platform_build functions. 
The purpose of AbsBuilder is to unbind the edk2_platform_build and UefiBuilder, so that edk2_platform_build can run other command, not only just edk2 build.py. This is the fundamental problem I need to resolve if I use Pytools.


> I think I would suggest you create your own invocable of the type want. I don't see value in this unless you get code reuse or simplification/scalability. If with each platform or new generation you are implementing a complete AbsBuilder then you haven't achieved those goals.

The AbsBuilder is the base class, I won't implement AbsBuilder, I will implement my builder class derive from the AbsBuilder.
I have thought about creating my own invocable and corresponding builder, that will introduce one more new command and there would be much similar code in the new invocable as it in edk2_platform_build.

BobCF:
This PR https://github.com/tianocore/edk2-pytool-extensions/pull/242 is about creating new invocable and builder.

Javagedes:
Closed due to inactivity, feel free to re-open!

PR #260, opened 2021-07-14
Add support for user overriding NuGet.exe from package using NUGET_PATH system environment variable.  Remove broken code that tried to find NuGet.exe on path.

spbrogan:
Quoted string fails os.path.isfile and thus a nuget.exe found on the path never works


corthon (approved):
Make sure the commit comment contains useful info about the change in functionality.

PR #261, opened 2021-07-15
Allow skipping multiple build steps from the cmdline

brandonaut:
This fixes an issue where if you try to skip multiple steps on the cmdline (e.g., `stuart_build --skipprebuild --skipbuild`), only one of the options would be honored. AFAICT, these options don't need to be mutually exclusive, so this PR allows them to be set independently.


spbrogan (approved):
Good catch.  
This wasn't ever supposed to be mutually exclusive but the most interesting part is this was ported from old code over 2 years ago and none of our users noticed.  I guess the most common scenario of SkipBuild would have worked so without log inspection it could have been missed.    Thanks again.

Issue #264, opened 2021-07-23
Omnicache fetch parallelism causes problems at scale.


joschock:
Omnicache 0.10 introduced parallelism for fetching. The implementation spawns a fetch job for each remote. On some memory-limited systems (such as build agent VMs), this can result in out-of memory issues when there are many remotes.

**To Reproduce**
1. Test Omnicache 0.10+ at scale
2. Observe fatal "out of memory" errors from git on memory-bound systems

**Expected behavior**
Reasonable parallelization.

Issue #263, opened 2021-07-23
Omnicache conversion does not proceed if a badly-formed ref is in the remote being converted.


joschock:
When converting from a pre-0.11 omnicache to an 0.11 omnicache, the omnicache utility will attempt to update the existing omnicache repo to an 0.11 state. This involves (among other things) renaming the remotes in the omnicache repo.

If one of these remotes is badly formed (e.g. casing issues on a case-aware but not case-sensitive), then the conversion process bails out, leaving the repo in a torn state. 

**To Reproduce**
1. Make a pre 0.11 Omnicache
2. create a mal-formed ref in the repo that will cause `git remote rename` to gack
3. attempt to convert to 0.11+ omnicache
4. See error, and conversion process aborts.

**Expected behavior**
conversion process skips (or removes) the offending remote and proceeds.

PR #265, opened 2021-07-23
Omnicache bugfixes

joschock:
Resolves a couple of issues discovered when testing omnicache 0.11 at scale:

1. #264 
2. #263

### DIFF
--- a/edk2toolext/omnicache.py
+++ b/edk2toolext/omnicache.py
@@ -16,7 +16,7 @@ from io import StringIO
 from edk2toolext import edk2_logging
 from edk2toollib.utility_functions import RunCmd
 
-# Omnicache version 0.11 design notes:
+# Omnicache version 0.11+ design notes:
 # 1. Only a single remote per URL will be permitted.
 # 2. Remote names inside omnicache will be UUIDs generated when a new URL is added.
 # 3. Users may specify names for remotes, but they will be treated simply as "display names" without git significance.
@@ -27,7 +27,7 @@ from edk2toollib.utility_functions import RunCmd
 #     older omnicache interacts with it. Older omnicache should not crash when encountering a newer omnicache directory
 #     but may take a lot of slow and unecessary actions to "re-initialize" it.
 
-OMNICACHE_VERSION = "0.11"
+OMNICACHE_VERSION = "0.12"
 
 # Prior to version 0.11, an "omnicache.yaml" file in the root of the omnicache repo was used to manage the cache
 # configuration. Starting with 0.11, git config entries are used directly. If the file below is present, it will be
@@ -95,6 +95,11 @@ class Omnicache():
         ret = RunCmd("git", "init --bare", workingdir=self.path)
         if (ret != 0):
             return ret
+
+        ret = RunCmd("git", "config --local fetch.parallel 0", workingdir=self.path)
+        if (ret != 0):
+            return ret
+
         return RunCmd("git",
                       "config --local omnicache.metadata.version {0}".format(OMNICACHE_VERSION),
                       workingdir=self.path)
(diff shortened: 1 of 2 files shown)


edk2toolext/omnicache.py
```diff
@@ -95,6 +95,11 @@ def _InitOmnicache(self):
         ret = RunCmd("git", "init --bare", workingdir=self.path)
         if (ret != 0):
             return ret
+
+        ret = RunCmd("git", "config --local fetch.parallel 0", workingdir=self.path)
```
corthon, line 99: Where are the details of these two config changes documented? Would it be appropriate to have a comment here about why the cache repo must be configured this way?

edk2toolext/omnicache.py
```diff
@@ -145,6 +155,10 @@ def _ConvertOmnicache(self):
                 logging.info("Removing remote {0} with duplicate URL {1}".format(name, url))
                 RunCmd("git", "remote remove {0}".format(name), workingdir=self.path)
                 RunCmd("git", "config --local --unset omnicache.{0}.displayname".format(name), workingdir=self.path)
+        # setup fetch.parallel to a reasonable default
```
corthon, line 158: Why is this default reasonable

edk2toolext/omnicache.py
```diff
@@ -393,6 +410,8 @@ def get_cli_options():
                        help="Update the Omnicache.  All cache changes also cause a fetch", default=False)
     group.add_argument("--no-fetch", dest="no_fetch", action="store_true",
                        help="Prevent auto-fetch if implied by other arguments.", default=False)
+    group.add_argument("--fetch-jobs", dest="fetch_jobs", type=int,
```
corthon, line 413: Is this the complement to "parallel" above?

corthon (approved):

edk2toolext/omnicache.py
joschock: Will add a comment; the default behavior of git fetch is single threaded. fetch.parallel = 0 uses "a reasonable default" (per the git docs) - I looked into the source and presently that means a number of threads equal to current cpu count. Original implementation I set up forked a thread for each remote; this was running our build agent VMs out of memory.

edk2toolext/omnicache.py
joschock: Will add a more informative comment; that was just the git verbiage.

edk2toolext/omnicache.py
joschock: Yes; it allows the user to override the "reasonable" default of a thread per CPU and force more or less parallelization if needed. I added it for maximum flexibility here; but most implementations will likely just use the default.

edk2toolext/omnicache.py
```diff
@@ -239,13 +255,16 @@ def UpdateRemote(self, oldUrl, newUrl=None, newName=None):
...
+    def Fetch(self, jobs=0):
         """Fetches all remotes"""
         logging.info("Fetching all remotes.")
         self._RefreshUrlLookupCache()
         # Tricky: we pass no-tags here, since we set up custom fetch refs for tags on a per-remote basis. This prevents
         # git from fetching the first set of tags into the global namespace.
-        return RunCmd("git", "fetch --all -j {0} --no-tags".format(len(self.urlLookupCache)), workingdir=self.path)
+        if (jobs != 0):
+            return RunCmd("git", "fetch --all -j {0} --no-tags".format(len(self.urlLookupCache)), workingdir=self.path)
+        else:
```
joschock, line 266: Bug: this should be the fetch-jobs argument. Will fix.

joschock: Fixed.

PR #272, opened 2021-08-19
Add initial support for Azure CLI universal packages as ext_dep type

spbrogan:
See https://docs.microsoft.com/en-us/azure/devops/artifacts/quickstarts/universal-packages?view=azure-devops for details on universal packages

### DIFF
--- a/docs/usability/using_extdep.md
+++ b/docs/usability/using_extdep.md
@@ -89,6 +89,22 @@ checked out. For this ext_dep descriptor the type is **_git_**. A git dependency
 should be treated as read-only because the verify and clean phase will do
 destructive operations where local changes would be destroyed.
 
+### Azure Universal Packages Dependency (az-universal)
+
+Azure Universal packages are yet another package management solution. These
+packages are supported by Azure DevOps server and use the cross platform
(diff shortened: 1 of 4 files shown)


spbrogan:
Unit Tests seem to fail on the server.  Looks like authentication related.  Az cli is not a very good scriptable interface so I am not sure this is going to work.

corthon (approved):

spbrogan:
@corthon  - there is no cache.  Should we create and manage our own file cache?

corthon:
Yeah, not sure what to do about the auth. Maybe some who have used them in existing pipelines have some ideas? I think this is one of the downsides about it not supporting public feeds (as far as I can tell).

spbrogan:
> Yeah, not sure what to do about the auth. Maybe some who have used them in existing pipelines have some ideas? I think this is one of the downsides about it not supporting public feeds (as far as I can tell).

working on a PAT based solution.  will see if it works

spbrogan:
The PAT solution works but it has lots of limitations because to depends on secret variables in the pipeline which are not allowed for PRs from forks.  I think it will work for actual product development which use devops / private repos and a development process  that leverages "user" branches for PRs in the repo rather than forks.

PR #277, opened 2021-10-27
Bug fix in merge_config

bkarstens:
Fixed a variable shadowing bug in `merge_config` where the `config` argument was lost.

### DIFF
--- a/edk2toolext/tests/test_edk2_ci_build.py
+++ b/edk2toolext/tests/test_edk2_ci_build.py
@@ -71,3 +71,37 @@ class TestEdk2CiBuild(unittest.TestCase):
             self.assertEqual(e.code, 0, "We should have a non zero error code")
             pass
         self.assertTrue(os.path.exists(os.path.join(self.minimalTree, "Build")))
+
+    def test_merge_config(self):
+        descriptor = {'descriptor_file': 'C:\\MyRepo\\.pytool\\Plugin\\MyPlugin1\\MyPlugin1_plug_in.yaml',
+                      'module': 'MyPlugin1',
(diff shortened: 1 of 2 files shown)


edk2toolext/tests/test_edk2_ci_build.py
```diff
@@ -71,3 +71,37 @@ def test_ci_build(self):
...
+        }
+        merged_config = {
+            "MySetting1": 'package value 1',
+            "MySetting2": 'global value 2',
+            "MySetting3": 'package value 3'
+        }
+
+        self.assertDictEqual(Edk2CiBuild.merge_config(global_config, {}, descriptor), global_config["MyPlugin1"])
+        self.assertDictEqual(Edk2CiBuild.merge_config({}, package_config, descriptor), package_config["MyPlugin1"])
+        self.assertDictEqual(Edk2CiBuild.merge_config(global_config, package_config, descriptor), merged_config)
```
spbrogan, line 107: Why does merged_config come out flattened and without MyPlugin2 or MyPlugin3?

bkarstens: The `merge_config` function is called on each plugin in `Go`. The descriptor object tells `merge_config` which plugin's settings to merge and return; in this test I've specified the plugin (module) in the descriptor to be `MyPlugin1`, so the merged settings for `MyPlugin1` are returned.

spbrogan: I forgot about the descriptor parameter.   Thanks

spbrogan (approved):

PR #282, opened 2022-02-08
Add ability to skip directories

makubacki:
Adds a GetSkippedDirectories() method that allows a SettingsManager
to specify a tuple of workspace-relative directories that should be
skipped.

This allows subdirectories (such as submodules) that might contain duplicate
resources that would normally cause build errors (such as external dependency
descriptors) to be ignored if only one directory is actually being used for
the current build.

### DIFF
--- a/edk2toolext/base_abstract_invocable.py
+++ b/edk2toolext/base_abstract_invocable.py
@@ -33,6 +33,10 @@ class BaseAbstractInvocable(object):
         ''' return tuple containing scopes that should be active for this process '''
         raise NotImplementedError()
 
+    def GetSkippedDirectories(self):
+        ''' return tuple containing workspace-relative directory paths that should be skipped for processing'''
+        return ()
+
     def GetLoggingLevel(self, loggerType):
         ''' Get the logging level for a given type (return Logging.Level)
         base == lowest logging level supported
@@ -114,11 +118,11 @@ class BaseAbstractInvocable(object):
         # Next, get the environment set up.
         #
         (build_env, shell_env) = self_describing_environment.BootstrapEnvironment(
-            self.GetWorkspaceRoot(), self.GetActiveScopes())
+            self.GetWorkspaceRoot(), self.GetActiveScopes(), self.GetSkippedDirectories())
 
         # Make sure the environment verifies IF it is required for this invocation
         if self.GetVerifyCheckRequired() and not self_describing_environment.VerifyEnvironment(
-                self.GetWorkspaceRoot(), self.GetActiveScopes()):
+                self.GetWorkspaceRoot(), self.GetActiveScopes(), self.GetSkippedDirectories()):
             raise RuntimeError("SDE is not current.  Please update your env before running this tool.")
 
         # Load plugins
--- a/edk2toolext/base_abstract_invocable.py
+++ b/edk2toolext/base_abstract_invocable.py
@@ -34,7 +34,8 @@ class BaseAbstractInvocable(object):
         raise NotImplementedError()
 
     def GetSkippedDirectories(self):
-        ''' return tuple containing workspace-relative directory paths that should be skipped for processing'''
+        ''' Return tuple containing workspace-relative directory paths that should be skipped for processing.
+        Absolute paths are not supported. '''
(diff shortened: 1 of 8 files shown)


edk2toolext/environment/self_describing_environment.py
```diff
@@ -46,8 +52,8 @@ def _gather_env_files(self, ext_strings, base_path):
         # matching the extension.
         matches = {}
         for root, dirs, files in os.walk(base_path, topdown=True):
-            # Check to see whether any of these directories should be skipped.
-            # TODO: Allow these to be passed in via arguments.
+            # Check to see whether any of these directories should be skipped..
+            dirs[:] = [d for d in dirs if pathlib.Path(root, d) not in self.skipped_dirs]
             for index, dir in enumerate(dirs):
```
spbrogan, line 57: Should the new path check be done inside the existing for loop?  Why do two iterations?

edk2toolext/environment/self_describing_environment.py
```diff
@@ -30,6 +33,9 @@ def __init__(self, workspace_path, scopes=()):
         # Start with the provided set.
         self.scopes = scopes
 
+        # Allow certain directories to be skipped
+        self.skipped_dirs = tuple(map(pathlib.Path, (os.path.join(self.workspace, d) for d in skipped_dirs)))
```
spbrogan, line 37: should it be only workspace related or should packages path be involved?

edk2toolext/edk2_invocable.py
```diff
@@ -61,6 +61,10 @@ def RetrieveCommandLineOptions(self, args: object) -> None:
         '''  Implement in subclass to retrieve command line options from the argparser namespace '''
         pass
 
+    def GetSkippedDirectories(self) -> Tuple[str]:
```
spbrogan, line 64: this is implemented in base_abstract_invocable class.  Why needed here for same implementation?

edk2toolext/base_abstract_invocable.py
```diff
@@ -33,6 +33,10 @@ def GetActiveScopes(self):
         ''' return tuple containing scopes that should be active for this process '''
         raise NotImplementedError()
 
+    def GetSkippedDirectories(self):
+        ''' return tuple containing workspace-relative directory paths that should be skipped for processing'''
```
spbrogan, line 37: should we support packages path too?

edk2toolext/invocables/edk2_update.py
```diff
@@ -61,6 +62,10 @@ def RetrieveCommandLineOptions(self, args):
         '''  Retrieve command line options from the argparser '''
         super().RetrieveCommandLineOptions(args)
 
+    def GetSkippedDirectories(self):
```
spbrogan, line 65: why is this needed if the super has the function?

edk2toolext/edk2_invocable.py
makubacki: I just forgot to remove after adding it in the abstract class. I'll update it to be removed.

edk2toolext/invocables/edk2_update.py
makubacki: Added when debugging something and I forgot to remove it. I'll update it.

edk2toolext/environment/self_describing_environment.py
makubacki: I was only considering workspace relative paths. Do you think packages path is needed?

spbrogan:
also note it looks like pr_eval on project mu is broken.  is that expected?

corthon (approved):
Approved with agreement with Sean.

makubacki:
> also note it looks like pr_eval on project mu is broken. is that expected?

Not expected, it was working in the previous push. It looks like GetSkippedDirectories() being removed from Edk2InvocableSettingsInterface() introduced it. I added it back to confirm.

edk2toolext/base_abstract_invocable.py
makubacki: I included a note about absolute paths.

spbrogan (approved):

PR #283, opened 2022-02-11
Create an ExtDep cache for any Stuart extdep type

corthon:
Needs some more documenting, but it seems to work well.
Also solves some of the problems that we've had with "malformed" Nuget cache.


spbrogan:
Can you add/update documentation.  
I assume both user and developer docs need a brief  update

spbrogan (approved):

corthon:
EDK2 integration test is known to be failing on public branches right now. Ran tests locally to verify.

Issue #294, opened 2022-04-07
Verify no nested package paths


Javagedes:
**Is your feature request related to a problem? Please describe.**
Nested package paths  ```pkg_paths = ["a/b/", "a/b/c"]``` is not supported by the build system due to undefined behavior. As an example nested pkg_paths can cause inconsistent behavior in edk2toollib>path_utilities>GetEdk2RelativePathFromAbsolutePath() as the generated relative path will be different depending on if it uses "a/b/" or "a/b/c" package path. This paired with Issue #293 means that you can have different results build to build. 

**Describe the solution you'd like**
Add build time verification to ensure no nested package paths. This can be easily done in:
https://github.com/tianocore/edk2-pytool-extensions/blob/3452e781a021e9b736fb10dbd3e7645a2efc400f/edk2toolext/edk2_invocable.py#L102

Javagedes:
This will be done in edk2-pytool-library

Issue #293, opened 2022-04-07
Package paths list order is different every build


Javagedes:
**Describe the bug**
Every time you build a platform, the order in which package paths are in the pack path list is different. It is expected that the order of package paths in the list remain the same from build to build, as the order is developer defined.

**To Reproduce**
1. Add logging information before and after https://github.com/tianocore/edk2-pytool-extensions/blob/3452e781a021e9b736fb10dbd3e7645a2efc400f/edk2toolext/edk2_invocable.py#L105
2. Note that the order of the package paths is different before and after this line, even if no directories are skipped.

**Expected behavior**
The order of the contents of the package path list should remain the same as defined by the developer, with the exception of any skipped directories.

PR #295, opened 2022-04-07
Update GetPackagesPath() to prevent package path list reordering

Javagedes:
This PR addresses #293 and #294 by updating how packages are removed so that the package path list is not reordered. It also adds in build time verification that there are no nested package paths.

### DIFF
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -102,19 +102,7 @@ class Edk2Invocable(BaseAbstractInvocable):
     def GetPackagesPath(self) -> Iterable[os.PathLike]:
         ''' Use the SettingsManager to an iterable of paths to be used as Edk2 Packages Path'''
         try:
-            packages = [x for x in self.PlatformSettings.GetPackagesPath() if x not in self.GetSkippedDirectories()]
-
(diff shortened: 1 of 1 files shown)


edk2toolext/edk2_invocable.py
```diff
@@ -102,7 +102,19 @@ def GetWorkspaceRoot(self) -> os.PathLike:
     def GetPackagesPath(self) -> Iterable[os.PathLike]:
         ''' Use the SettingsManager to an iterable of paths to be used as Edk2 Packages Path'''
         try:
-            return list(set(self.PlatformSettings.GetPackagesPath()) - set(self.GetSkippedDirectories()))
+            packages = [x for x in self.PlatformSettings.GetPackagesPath() if x not in self.GetSkippedDirectories()]
+
+            # Temporarily sort, remove duplicates, and match os specific
```
spbrogan, line 107: can we move this into its own function?  Should this "check for nesting" be in edk2-pytool-library?   Maybe https://github.com/tianocore/edk2-pytool-library/blob/3de82123f055e65c3babbc4e5833121f4215af51/edk2toollib/uefi/edk2/path_utilities.py#L56

spbrogan (approved):

PR #298, opened 2022-04-07
Update azure windows vm and corresponding tool chain for integration pipelines

Javagedes:
Update EDK2 and Project Mu Integration pipelines to the following:
Edk2: windows-2019 VM, VS2019 tool chain
Project Mu: windows-2022 VM, VS2022 tool chain

### DIFF
--- a/integration_test/azure-pipelines/windows-robot-integration-test.yml
+++ b/integration_test/azure-pipelines/windows-robot-integration-test.yml
@@ -15,13 +15,13 @@ jobs:
     matrix:
       PROJECT_MU:
         Tag: 'ProjectMu'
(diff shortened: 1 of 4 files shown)


makubacki (approved):

integration_test/azure-pipelines/windows-robot-integration-test.yml
```diff
@@ -18,10 +18,10 @@ jobs:
         Image: 'windows-latest'
```
spbrogan, line 18: I would suggest making this windows-2022 so that we are always locked in to a version.  
We do this now for mu ci as well.

spbrogan (approved):

PR #305, opened 2022-08-02
Fix CI Failures

Javagedes:
Fix flake8 E275 failures which requires a space after a keyword (in particular, the **assert**, **if**, **elif**, and **else** keywords)
Fix integration test failures due to recent changes in both Project Mu and EDK2

### DIFF
--- a/BasicDevTests.py
+++ b/BasicDevTests.py
@@ -79,15 +79,15 @@ py_files = glob.glob(os.path.join(p, "**", "*.py"), recursive=True)
 error = 0
 for a in py_files:
     aRelativePath = os.path.relpath(a, os.getcwd())
-    if(not TestEncodingOk(a, "ascii")):
+    if (not TestEncodingOk(a, "ascii")):
         error += 1
-    if(not TestFilenameLowercase(aRelativePath)):
+    if (not TestFilenameLowercase(aRelativePath)):
         error += 1
-    if(not TestNoSpaces(aRelativePath)):
+    if (not TestNoSpaces(aRelativePath)):
         error += 1
-    if(not TestRequiredLicense(a)):
+    if (not TestRequiredLicense(a)):
         error += 1
-    if(not PackageAndModuleValidCharacters(aRelativePath)):  # use relative path so only test within package
+    if (not PackageAndModuleValidCharacters(aRelativePath)):  # use relative path so only test within package
         error += 1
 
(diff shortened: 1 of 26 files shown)


makubacki:
Since this is just a style change (or should be), can you also get rid of parentheses around conditional expressions after keywords like "if"?

That's a C style that's non-Pythonic and creates a style debt that doesn't align with new code.

Javagedes:
> Since this is just a style change (or should be), can you also get rid of parentheses around conditional expressions after keywords like "if"?
> 
> That's a C style that's non-Pythonic and creates a style debt that doesn't align with new code.

@makubacki This is a style change, but this PR was made because it is a new flake8 rule that we are failing which is causing us to fail CI. While I can remove the parentheses manually, it is not a Flake8 rule, so with the current infrastructure, it cannot be maintained / enforced for future PRs. 

One option I see is to require compliance with both flake8 and pylint. They have a lot of overlap, but also some differences (such as pylint having the parenthesis requirement you're looking for [C0325](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/superfluous-parens.html)).  If you think it is useful to go down this route, we could also throw in [bandit](https://pypi.org/project/bandit/) on the security side too.

Thoughts @makubacki @spbrogan ?

makubacki:
> 

I suggest we get this in for now to resolve the immediate problem and then look at ways to converge linting like you suggested.

makubacki:
This PR has a few unrelated changes. The commits are currently prepared for a squash operation.

Can you please break them out to individual commits and the Rebase & FF the series?

makubacki:
> This PR has a few unrelated changes. The commits are currently prepared for a squash operation.
> 
> Can you please break them out to individual commits and the Rebase & FF the series?

Mostly looks like (1) space change across all files (2) using.md broken link fix (3) some .robot file changes.

makubacki (approved):
Please help address the few minor comments I left before completing.

Javagedes:
> > This PR has a few unrelated changes. The commits are currently prepared for a squash operation.
> > Can you please break them out to individual commits and the Rebase & FF the series?
> 
> Mostly looks like (1) space change across all files (2) using.md broken link fix (3) some .robot file changes.

I have rebased the commits to represent the three major changes and will rebase & FF when merging. Thanks!

cfernald (approved):

PR #308, opened 2022-08-29
docs/using.md: Fix broken internal document link

makubacki:
[stuart commands](#what-can-ask-stuart-to-do)

should be

[stuart commands](#what-can-i-ask-stuart-to-do)


Javagedes:
This change is already in #305. Tests will continue to fail until edk2-pytool-library has a release as there is a necessary fix there.

PR #316, opened 2022-09-16
Improve NuGet dependency semantic version handling

makubacki:
NuGet package versioning is described here:
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning

It follows semantic versioning described here:
https://semver.org/

Among other things, semantic versioning allows an alphanumeric pre-release version number:
https://semver.org/#spec-item-9

The current NuGet versioning normalization code in nuget_dependency.py only permits a limited set of pre-approved tags such as the following to serve as the pre-release version number:

  ["beta", "alpha", "rc"]

Therefore, a valid NuGet package version string such as the following is considered invalid:

  "2.8.4-92" -[reports]-> "ValueError: Unparsable version tag: 92"

In other cases, the previous logic allowed passed invalid versions. For example:

  "2.15.05"

Which should not be allowed to use a leading zero: https://semver.org/#spec-item-2

This change swaps out the custom semantic versioning handling logic with more formal support provided by the semantic_version module which does correctly validate this string and has better compatibility with the Semantic Versioning Specification.

The "normalize_version()" static method is now made a class method so it can have access to object data fields. There was not a strong use case found for its static ability.

A clear message identifying the NuGet dependency and invalid version string is printed in case of an error.

In addition, "test_normalize_version()" is removed from test_nuget_dependency.py since there is no longer custom logic to validate.

### DIFF
--- a/edk2toolext/environment/extdeptypes/nuget_dependency.py
+++ b/edk2toolext/environment/extdeptypes/nuget_dependency.py
@@ -8,6 +8,7 @@
 ##
 import os
 import logging
+import semantic_version
 import shutil
 from io import StringIO
 from edk2toolext.environment.external_dependency import ExternalDependency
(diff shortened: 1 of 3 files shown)


makubacki:
Note: I expect some integration tests to fail. For example, `mu_nasm` has an invalid version `2.15.05` due to the leading zero.

makubacki:
> Note: I expect some integration tests to fail. For example, `mu_nasm` has an invalid version `2.15.05` due to the leading zero.

https://github.com/tianocore/edk2/blob/master/BaseTools/Bin/nasm_ext_dep.yaml#L16

Javagedes:
Looks good to me. Integration tests are failing on the Project Mu side, so easy fix. Once I get that in, I'll approve.

makubacki:
> > Note: I expect some integration tests to fail. For example, `mu_nasm` has an invalid version `2.15.05` due to the leading zero.
> 
> https://github.com/tianocore/edk2/blob/master/BaseTools/Bin/nasm_ext_dep.yaml#L16

Change to remove the leading zero is out: https://edk2.groups.io/g/devel/message/93913

makubacki:
> Looks good to me. Integration tests are failing on the Project Mu side, so easy fix. Once I get that in, I'll approve.

mu_basecore fixes are in - https://github.com/microsoft/mu_basecore/commit/ea9bd5f8f46b8ff2ff1a42393210da96cde07cb4

Javagedes (approved):

makubacki:
@Javagedes & @spbrogan - Since this is approved, is there anything else needed for it to be merged?

Javagedes:
> @Javagedes & @spbrogan - Since this is approved, is there anything else needed for it to be merged?

No, you can merge this at any time. I'll still need to do a release and update any pip-requirements for the change to take effect, however. I don't plan on doing a release for just this one single change, unless its urgent.

makubacki:
> No, you can merge this at any time. I'll still need to do a release and update any pip-requirements for the change to take effect, however. I don't plan on doing a release for just this one single change, unless its urgent.

A separate release is not necessary right now.

makubacki:
> > No, you can merge this at any time. I'll still need to do a release and update any pip-requirements for the change to take effect, however. I don't plan on doing a release for just this one single change, unless its urgent.
> 
> A separate release is not necessary right now.

Also, I don't have write access to the repo so I can't merge. Can you please merge it?

PR #317, opened 2022-09-27
Bug fix: bad variable name caused a trace in an error branch

bkarstens:


Javagedes:
@bkarstens please add a corresponding test case to [test_edk2_setup.py](https://github.com/tianocore/edk2-pytool-extensions/blob/master/edk2toolext/tests/test_edk2_setup.py) to ensure this bug does not occur again... Similar to https://github.com/tianocore/edk2-pytool-extensions/pull/315

Javagedes (approved):

PR #324, opened 2022-10-31
Add CodeQL build support

makubacki:
Adds the ability to generate a CodeQL database during build.

The tracing needed to build the CodeQL database must wrap the `build`
command which leads to the changes being concentrated in
uefi_build.py.

By default, the changes focus on adding support to UefiBuilder
subclasses (i.e. platform builders via `stuart_build`) via two new
command line parameters:

  1. `--codeql` - Produce CodeQL database from build. Implies
     `--clean`.
     - A clean build is required for CodeQL to generate a database
       properly.

  2. `--codeqldbpath` - Path relative to workspace root where the
     CodeQL database is created.

In addition, those that interact with UefiBuilders outside of
inheritance (such as build plugins) can enable CodeQL functionality
in a UefiBuilder object via the `EnableCodeQl()` method.

The user is expected to run additional tools outside of the main
Stuart build process to generate CodeQL results from the generated
database.

### DIFF
--- a/edk2toolext/environment/uefi_build.py
+++ b/edk2toolext/environment/uefi_build.py
@@ -11,6 +11,8 @@
 
 import os
 import logging
+import shutil
+import stat
 from edk2toolext.environment.multiple_workspace import MultipleWorkspace
 from edk2toolext.environment import conf_mgmt
 import traceback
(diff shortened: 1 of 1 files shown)


makubacki:
After further thought, I've decided to make CodeQL automation available outside the stuart tools. This will make the scripts more generic for those not using stuart and avoid potential duplication with server execution happening through the GitHub action and code scanning workflow.

PR #327, opened 2022-11-01
edk2_pr_eval.py: Build all packages on file change outside package

makubacki:
Updates `get_packages_to_build()` to return all possible packages if
a file is modified outside a package.

Originally, this was implemented behind a new argument with the
default behavior unchanged. However, feedback indicated it would
be easier to adopt without an argument and this would be acceptable
behavior.

This does mean some files, such as documentation files, could trigger
packages to build that would have not previously built. However, the
impact and frequency of this is considered minimal enough to not
justify additional complexity. The change mainly focuses on files
such as dependency trackers (e.g. PIP requirements, submodules),
Python scripts, and other files that can impact overall build
results.


makubacki:
I'm working on the integration tests now.

makubacki:
> I'm working on the integration tests now.

Updated. As noted in the commit message, the tests will not pass until https://github.com/tianocore/edk2-pytool-library/pull/186 is available.

edk2toolext/invocables/edk2_pr_eval.py
```diff
@@ -120,6 +120,22 @@ def get_packages_to_build(self, possible_packages: list) -> dict:
         if rc != 0:
             return {}
 
+        #
+        # Policy 0 - A file outside a package was modified and force build
+        #            all packages on a modified file outside a package setting
+        #            is enabled.
+        #
+        # Note: GetContainingPackage() returns the top most directory from the
```
spbrogan, line 128: this comment is wrong now, right?

makubacki: Yeah, I was waiting for the edk2-pytool-library change to get unblocked to come back to this since it's technically not until the PR over there gets in.

makubacki: Updated

makubacki:
I preemptively moved the `edk2-pytool-library` dependency version to 0.12.1 (which doesn't exist yet). This should be completed after that release is done.

makubacki:
> I preemptively moved the `edk2-pytool-library` dependency version to 0.12.1 (which doesn't exist yet). This should be completed after that release is done.

Because of that, I added a distinct commit to show the setup.py update. Please Rebase & FF to keep that separate.

Javagedes:
release pipeline is running now and should be out soon.

Javagedes (approved):

PR #339, opened 2022-11-07
Plugin Loader Updates [Rebase & FF]

### LINKED ISSUE #337: Plugin Loader dependency will be removed in Python 3.12
### Contact Details

_No response_

### Describe the Bug

  Plugins are currently loaded in the `_load()` function in
  plugin_manager.py by using the `imp` standard library:

  https://docs.python.org/3/library/imp.html

  The first sentence of the library documentation states the following:

  ```
  Deprecated since version 3.4, will be removed in version 3.12: The
  imp module is deprecated in favor of importlib.
  ```

The code should be updated to use importlib.

### Reproduction steps

1. Review code in plugin_manager.py
2. Read https://docs.python.org/3/library/imp.html

### Expected behavior

Use importlib to prevent breaking change in the Python 3.12 release.

### Execution Environment

_No response_

### Pip packages

_No response_

### Additional context

_No response_

### LINKED ISSUE #338: [Feature]: Allow plugins to reference relative modules
### What does the feature solve?

Today, plugins cannot resolve relative modules because they are dynamically loaded, and their path is not in Python's search paths.

This is inconsistent with the import experience of Python modules and confuses IDEs when they validate import paths. The IDEs recognize a module should be capable of being imported successfully that is located relative to the imported module, but it does not work at runtime.

### Describe the solution

Add the directory path to the plugin being imported to the system path so the plugin can import modules relative to that path.

### Have you considered any alternatives?

_No response_

### Additional context

_No response_

makubacki:
This series makes two main changes:

1. [Update usage of deprecated loading APIs](https://github.com/tianocore/edk2-pytool-extensions/commit/49a954c5153d0e8acba8bca88e1b8ef962acd885)

   Fixes #337

   Plugins are currently loaded in the `_load()` function in
   plugin_manager.py by using the `imp` standard library:

   https://docs.python.org/3/library/imp.html

   The first sentence of the library documentation states the following:

   ```
   Deprecated since version 3.4, will be removed in version 3.12: The
   imp module is deprecated in favor of importlib.
   ```

   This means, `imp` will be removed in the next Python release.

   Therefore, this change updates the code to use `importlib` standard
   library as recommended by the `imp` documentation.

   The code follows the recipe recommended by the `importlib`
   documentation to directly load a Python source file as is the case
   when loading Python modules based on plugin descriptor information.

   https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

2. [Add plugin module paths to system path](https://github.com/tianocore/edk2-pytool-extensions/commit/a5757ddf73eb8034a346aac26a73082ae63964be)

   Closes #338 

   Adds dynamically loaded plugin module directory paths to the system
   path so the plugin modules can import other modules relative to their
   path as they would if they were not dynamically loaded.

### DIFF
--- a/edk2toolext/environment/plugin_manager.py
+++ b/edk2toolext/environment/plugin_manager.py
@@ -95,6 +95,11 @@ class PluginManager(object):
                 py_module_name, py_module_path)
             module = importlib.util.module_from_spec(spec)
             sys.modules[py_module_name] = module
+
+            py_module_dir = os.path.dirname(py_module_path)
+            if py_module_dir not in sys.path:
+                sys.path.append(py_module_dir)
+
             spec.loader.exec_module(module)
         except Exception:
             exc_info = sys.exc_info()
--- a/edk2toolext/environment/plugin_manager.py
+++ b/edk2toolext/environment/plugin_manager.py
@@ -9,7 +9,7 @@
 
 import sys
 import os
-import imp
+import importlib
 import logging
 from edk2toolext.environment import shell_environment
 
@@ -82,32 +82,34 @@ class PluginManager(object):
             PluginDescriptor(PluginDescriptor): the plugin descriptor
         """
         PluginDescriptor.Obj = None
-        PythonFileName = PluginDescriptor.descriptor["module"] + ".py"
-        PyModulePath = os.path.join(os.path.dirname(os.path.abspath(
-            PluginDescriptor.descriptor["descriptor_file"])), PythonFileName)
-        logging.debug("Loading Plugin from %s", PyModulePath)
-        try:
-            with open(PyModulePath, "r") as plugin_file:
-                _module = imp.load_module(
-                    "UefiBuild_Plugin_" + PluginDescriptor.descriptor["module"],
-                    plugin_file,
-                    PyModulePath,
-                    ("py", "r", imp.PY_SOURCE))
 
+        py_file_path = PluginDescriptor.descriptor["module"] + ".py"
+        py_module_path = os.path.join(os.path.dirname(os.path.abspath(
+            PluginDescriptor.descriptor["descriptor_file"])), py_file_path)
+        py_module_name = "UefiBuild_Plugin_" + PluginDescriptor.descriptor["module"]
+
+        logging.debug("Loading Plugin from %s", py_module_path)
+
+        try:
+            spec = importlib.util.spec_from_file_location(
+                py_module_name, py_module_path)
+            module = importlib.util.module_from_spec(spec)
+            sys.modules[py_module_name] = module
+            spec.loader.exec_module(module)
         except Exception:
             exc_info = sys.exc_info()
             logging.error("Failed to import plugin: %s",
-                          PyModulePath, exc_info=exc_info)
+                          py_module_path, exc_info=exc_info)
             return -1
 
         # Instantiate the plugin
         try:
-            obj = getattr(_module, PluginDescriptor.descriptor["module"])
+            obj = getattr(module, PluginDescriptor.descriptor["module"])
             PluginDescriptor.Obj = obj()
         except AttributeError:
             exc_info = sys.exc_info()
             logging.error("Failed to instantiate plugin: %s",
-                          PyModulePath, exc_info=exc_info)
+                          py_module_path, exc_info=exc_info)
             return -1
 
         return 0


edk2toolext/environment/plugin_manager.py
```diff
@@ -82,32 +82,34 @@ def _load(self, PluginDescriptor):
...
-                    plugin_file,
-                    PyModulePath,
-                    ("py", "r", imp.PY_SOURCE))
 
+        py_file_path = PluginDescriptor.descriptor["module"] + ".py"
+        py_module_path = os.path.join(os.path.dirname(os.path.abspath(
+            PluginDescriptor.descriptor["descriptor_file"])), py_file_path)
+        py_module_name = "UefiBuild_Plugin_" + PluginDescriptor.descriptor["module"]
+
+        logging.debug("Loading Plugin from %s", py_module_path)
```
Javagedes, line 91: `logging.debug(f'Loading Plugin from {py_module_path})`

I find this way of doing variable insertion into strings much more readable. Not forcing you to make the change though.

edk2toolext/environment/plugin_manager.py
```diff
@@ -82,32 +82,34 @@ def _load(self, PluginDescriptor):
...
+
+        try:
+            spec = importlib.util.spec_from_file_location(
+                py_module_name, py_module_path)
+            module = importlib.util.module_from_spec(spec)
+            sys.modules[py_module_name] = module
+            spec.loader.exec_module(module)
         except Exception:
             exc_info = sys.exc_info()
             logging.error("Failed to import plugin: %s",
```
Javagedes, line 102: `logging.error(f'Failed to import plugin: {py_module_path}, exc_info=exc_info)`

Same as the previous comment. Optional to change but I find it easier to read (mainly when there are multiple variables being inserted into a string).

edk2toolext/environment/plugin_manager.py
```diff
@@ -82,32 +82,34 @@ def _load(self, PluginDescriptor):
...
             return -1
 
         # Instantiate the plugin
         try:
-            obj = getattr(_module, PluginDescriptor.descriptor["module"])
+            obj = getattr(module, PluginDescriptor.descriptor["module"])
             PluginDescriptor.Obj = obj()
         except AttributeError:
             exc_info = sys.exc_info()
             logging.error("Failed to instantiate plugin: %s",
```
Javagedes, line 112: `logging.error(f'Failed to instantiate plugin: {py_module_path}', exc_info=exc_info)`

edk2toolext/environment/plugin_manager.py
```diff
@@ -82,32 +82,35 @@ def _load(self, PluginDescriptor):
...
+        py_module_name = "UefiBuild_Plugin_" + PluginDescriptor.descriptor["module"]
+
+        logging.debug("Loading Plugin from %s", py_module_path)
+
+        try:
+            spec = importlib.util.spec_from_file_location(
+                py_module_name, py_module_path)
+            module = importlib.util.module_from_spec(spec)
+            sys.modules[py_module_name] = module
+            sys.path.append(os.path.dirname(py_module_path))
```
spbrogan, line 98: is it required to add the folder containing the module to the python path?  
This will impact all future loads of modules, right?

spbrogan (approved):
approve with comment about polluting the python path and potential unintended consequences.

edk2toolext/environment/plugin_manager.py
makubacki: I do too but that comment is just left as-is from before (apart from the variable name refactor). With many of these refactor changes, there's eventually a line as to whether it is better to fix isolated cases like this or go through the whole codebase and use f-strings.

I decided not to cross the line in this case since this is not a new `logging.debug` addition to the codebase and the previous style is not at risk of being deprecated/removed (at least to my knowledge).

Javagedes: Fair enough. I’ll raise an issue to go through and replace them all.

edk2toolext/environment/plugin_manager.py
makubacki: Like the other, I'll leave this to the f-string refactor

edk2toolext/environment/plugin_manager.py
Javagedes: `sys.path.append(os.path.dirname(py_module_path))` is the point of contention.

I don't want to speak for Sean, but he was either asking: 

1. Is it necessary to append this path to PYTHONPATH, the os environment, not just to sys.path
or
 2. He thinks appending to sys.path also appends to PYTHONPATH and he was asking the consequences of doing an unchecked append and polluting PYTHONPATH with the same paths from running a build over and over.
 
 If the concern is 1... It is not necessary to also append to PYTHONPATH as the plugins are loaded on every build, so sys.path variable will be freshly generated and loaded with the module paths each build.

If the concern is 2... sys.path is freshly built on every build from PYTHONPATH, however updates to sys.path are not mirrored to PYTHONPATH, so we don't have to worry about the appends to sys.path polluting PYTHONPATH.

The only slight concern is that if you have multiple plugins in the same directory, you'll append that directory multiple times.

@spbrogan let me know if I'm off the mark.

makubacki: Right, the behavior of the `sys.path` list is described in its documentation:
https://docs.python.org/3/library/sys.html#sys.path

For this:

> The only slight concern is that if you have multiple plugins in the same directory, you'll append that directory multiple times.

It is not different than what plugins previously did - If they understood what was wrong, they would have to `sys.path.append()` their module path before their import. Same problem, two+ plugins in the same directory, would append the same path.

But I agree that it doesn't make sense to push the same path twice. I updated to only append if needed:

```python
            if py_module_dir not in sys.path:
                sys.path.append(py_module_dir)
```

PR #353, opened 2022-12-01
Run Integration Tests on last three python minor versions

Javagedes:
Updates the Integration Test pipeline to run on the last three python minor versions rather then only the most current minor version.

Signed-off-by: Joey Vagedes <joeyvagedes@microsoft.com>


integration_test/azure-pipelines/windows-robot-integration-test.yml
```diff
@@ -7,21 +7,34 @@
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
 
+parameters:
+- name: PythonVersionList
+  type: object
+  default:
+    # Job Titles cannot have periods, so we iterate
+    # With a map using entry.key and entry.value
+    '3_9': '3.9'
```
makubacki, line 16: Can you define these in one place?

There are multiple options:

- Define in variables / variable group in the ADO pipeline UI
- Define in a single file and bring it in as a template
- Etc.

makubacki (approved):
One feedback item to consolidate N-3 versions. Looks fine otherwise.

integration_test/azure-pipelines/windows-robot-integration-test.yml
Javagedes: As we are only running integration tests on the pipeline, there is only one location on the pipeline yaml to change and one location on the github action to change. resolving this as won't fix.

makubacki:
Thanks for updating to ubuntu 20.04 & 22.04. Looks good to me.

PR #356, opened 2022-12-06
Set close_fds=False when building

jgarver:
This change adds close_fds=False to RunCmd when launching the build.

This is necessary to enable GNU make's job control to pass through. GNU make's job control is implemented using tokens on file descriptor.  If the file descriptors remain open, a GNU make calling stuart_build can communicate with a GNU make called by stuart_build.  However, Python's Popen closes file descriptors by default.

A similar change will be proposed in edk2's build.py, completing the connection from the calling make to the called make.

### DIFF
--- a/edk2toolext/environment/uefi_build.py
+++ b/edk2toolext/environment/uefi_build.py
@@ -4,6 +4,7 @@
 # more extensive and custom behavior.
 ##
 # Copyright (c) Microsoft Corporation
+# Copyright (c) 2022 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
@@ -321,7 +322,7 @@ class UefiBuilder(object):
             edk2_build_params = params
         logging.debug(f"Edk2 build parameters are {edk2_build_params}")
 
-        ret = RunCmd(edk2_build_cmd, edk2_build_params)
+        ret = RunCmd(edk2_build_cmd, edk2_build_params, close_fds=False)
         # WORKAROUND - Undo the workaround.
         env.restore_checkpoint(pre_build_env_chk)
 


Javagedes:
@jgarver Thanks for submitting this! Can you explain the issue a little clearer? Is this an issue with calling stuart_build directly on the command line, or are you calling stuart_build from a make file?

I'm not overly concerned with this change as it passes all our CI, meaning stuart_build still works correctly on both linux and windows ... but I still want to understand the issue a little better, and replicate it, before merging.

Thanks!

jgarver:
Thanks for your consideration, @Javagedes!

This is only an issue when calling stuart_build from GNU make.  Our internal system wraps stuart, as I suspect many people end up doing at least for automated builds.

make is usually run as "make -j N" where N is the number of parallel jobs allowed.  It's also common for make to call make. In that case, you want the top make and the submake to coordinate with each other to make sure total number of jobs is no more than N.  This is the "job control" I mentioned and GNU make pulls this off using file descriptors.  The top make opens a pipe to the submake and they pass around job tokens.  This works great as long as make is calling make directly. Things get more interesting when there's a program or script in between.

In our case, we have make calling stuart calling build.py calling make.  We can tell the top make "trust us, there's a submake, open your job control file descriptor," but we need the guys in between to leave the file descriptor open.

In python 3.x, they changed the behavior of Popen such that it will close file descriptors by default before forking.  This is generally considered good for security.  But it's bad for make's job control.

Setup to replicate (written with Linux in mind):
```
git clone  https://github.com/tianocore/edk2.git
cd edk2
stuart_update -c ArmVirtPkg/PlatformCI/PlatformBuild.py -a AARCH64 TOOL_CHAIN_TAG=GCC5
stuart_setup -c ArmVirtPkg/PlatformCI/PlatformBuild.py -a AARCH64 TOOL_CHAIN_TAG=GCC5
python3 BaseTools/Edk2ToolsBuild.py -t GCC5
```

Add a Makefile at the top level with the contents:
```
build:
        +stuart_build -c ArmVirtPkg/PlatformCI/PlatformBuild.py -a AARCH64 TOOL_CHAIN_TAG=GCC5
```

Then build it (assuming you have GNU make in path):
```
make -j 4 build
```

In the output, you'll see this warning:
```
 make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```

That's the submake noting that it cannot talk to the top make, so it's defaulting to a single job at a time and our CPU cores lay idle.

Unfortunately, even with this PR applied, you'll still see this message.  That's because we still need to make a similar change in build.py in edk2.  I'll propose that change after this one is in.

BTW, another way to remove this warning is to unset the MAKEFLAGS env before calling stuart_build.  Then the submake will not be aware the top make exists and will not attempt job control.  Unfortunately, that becomes a problem when multiple stuart_build's are run in parallel, which is something we want to do in our build system to maximize CPU utilization.  Since there's no coordination between the submakes, they try to fork bomb the build server.

Thanks again for your time!

spbrogan (approved):

jgarver:
Fantastic.  Thanks!

PR #358, opened 2022-12-08
Transition CI from azure pipeline to Github Action

Javagedes:
This PR replaces much of the CI performed by the PR gate azure pipeline with a github action that does the same. 

This github action does the following:
* Run Unittests on the 3 specified python versions (latest 3) and on ubuntu-latest & windows-latest
* Run CI tests such as flake8, codecov, etc, on using ubuntu-latest and the specified python version (latest)


.github/workflows/prgate.yml
```diff
@@ -0,0 +1,95 @@
...
+      
+      - name: Upload coverage to codecov
+        uses: codecov/codecov-action@v2
+        with:
+          verbose: false
+
+      - name: Install npm dependencies
+        run: |
+          npm install -g markdownlint-cli
+          npm install -g cspell@5.20.0
```
spbrogan, line 79: why is this version defined here.  But not for markdownlint-cli? And Flake8 uses a version stored in env.  should probably lock versions for all things.

.github/workflows/prgate.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Install pip dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install -r docs/user/requirements.txt
+          pip install -e .
+
+      - name: Run code coverage
+        run: |
+          pip install coverage
```
spbrogan, line 68: why isn't this in a requirements file and strongly versioned?

.github/workflows/prgate.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.node-version}}
+      
+      - name: Install pip dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install -r docs/user/requirements.txt
+          pip install -e .
```
spbrogan, line 64: why no --upgrade to force the version

.github/workflows/prgate.yml
Javagedes: Added to the requirements file and strongly versioned for dependabot rather than installing the latest here.

.github/workflows/prgate.yml
Javagedes: Flake8 is a pip module installed (and strongly versioned) in the requirements file. I've locked the version for markdownlint-cli.

.github/workflows/prgate.yml
Javagedes: Added upgrade

readme.md
```diff
@@ -25,20 +29,27 @@ This is a supplemental package and is not required to be used for edk2 builds.
...
-| Linux Ubuntu 20.04 | Python 3.10 | Project Mu | [![Build Status](https://dev.azure.com/tianocore/edk2-pytool-extensions/_apis/build/status/Master%20CI%20Build%20-%20Linux?branchName=master)](https://dev.azure.com/tianocore/edk2-pytool-extensions/_build/latest?definitionId=7&branchName=master) | ![mu-ubuntu-test2] | ![Azure DevOps coverage](https://img.shields.io/azure-devops/coverage/tianocore/edk2-pytool-extensions/7.svg) |
-| Linux Ubuntu 20.04 | Python 3.11 | Project Mu | [![Build Status](https://dev.azure.com/tianocore/edk2-pytool-extensions/_apis/build/status/Master%20CI%20Build%20-%20Linux?branchName=master)](https://dev.azure.com/tianocore/edk2-pytool-extensions/_build/latest?definitionId=7&branchName=master) | ![mu-ubuntu-test3] | ![Azure DevOps coverage](https://img.shields.io/azure-devops/coverage/tianocore/edk2-pytool-extensions/7.svg) |
+[![codecov]][_codecov]
+[![ci]][_ci]
+
+The code coverage and CI badges represent unit test status and the code
+coverage of those unit tests. We require 100% unit test success
+(Hence the pass / fail), and 70% code coverage.
```
makubacki, line 37: May want to consider making the code coverage value more generic to prevent updates/confusion in the future when it changes.

Something like "... and lower percentage code coverage."

.github/workflows/ci.yml
```diff
@@ -0,0 +1,110 @@
...
+      with:
+        verbose: false
+    
+    - uses: actions/upload-artifact@v3
+      with:
+        name: Artifacts-${{matrix.os}}-${{matrix.python-version}}
+        path: |
+          pytest_report.html
+          test.junit.xml
+      if: failure()
```
makubacki, line 56: Is this meant to only upload the artifacts on failure?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,110 @@
...
+  
+  ci:
+    name: CI Tests
+    runs-on: ubuntu-latest
+    env:
+      python-version: "3.11"
```
makubacki, line 62: The idea is that the tests should actually be run with most recent Python version supported, right?

It would be nice to do something like get the max Python version from the matrix array and use it here. Not sure what that would look like doing something like defining the values outside the `unit-test` job and then referencing them in the matrix and here.

But, maybe at least a comment around Line 23 that says to bump this value based on the new max would be helpful.

makubacki (approved):
A few minor comments, looks good overall.

.github/workflows/ci.yml
Javagedes: Solved. Unfortunately, the strategy matrix does not have access to env variables to do this easily. Instead, I had to create an initial job that runs that the CI tests and Unit Tests depends on. This initial job generates an output that the other two jobs are able to use. The unit tests use the entire list of python versions to generate the matrix. The CI Tests only accesses the first index.

https://github.com/tianocore/edk2-pytool-extensions/pull/358/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR15

readme.md
Javagedes: Updated.

PR #364, opened 2022-12-12
Add Documentation Deployment Action

Javagedes:
Adds a github action that executes on a release that deploys the documentation to the pages site.

Example Execution on fork:

Action: https://github.com/Javagedes/edk2-pytool-extensions/actions/workflows/doc-release.yml

Site: https://javagedes.github.io/edk2-pytool-extensions/

### DIFF
--- a/.github/workflows/doc-release.yml
+++ b/.github/workflows/doc-release.yml
@@ -1,0 +1,62 @@
+# This workflow builds and deploys documentation on a release.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+
+name: Doc Deployment
+
+on:
+  release:
+    types: [published]
+
+
+jobs:
+  build:
+    name: Build and Upload Site Artifact
+    
+    runs-on: ubuntu-latest
+
+    env:
+      python-version: "3.11"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      
+      - name: Set up Python ${{ env.python-version }}
+        uses: actions/setup-python@v3
+        with:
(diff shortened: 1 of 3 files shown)


docs/user/requirements.txt
```diff
@@ -1,3 +1,4 @@
+black==22.12.0
```
makubacki, line 1: Is this actually needed for the site publishing changes?

.github/workflows/doc-release.yml
```diff
@@ -0,0 +1,62 @@
...
+      
+      - name: Build Documentation
+        run: |
+          mkdocs build --strict
+      
+      - name: Upload Github Pages Site Artifact
+        uses: actions/upload-pages-artifact@v1
+        with:
+          name: 'github-pages'
+          path: "site/"
```
makubacki, line 44: Is there a pattern to the different quote styles being used? You don't need to change it, just curious why two simple strings close together have single and double quotes.

docs/user/requirements.txt
makubacki: I noticed it is in #362 as well. I think it makes more sense to get in there.

makubacki (approved):

docs/user/requirements.txt
Javagedes: Sounds good I'll merge that Pr first and remove it from here.

Black is used for some additional checks within mkdocs when building.

Javagedes: Removed

.github/workflows/doc-release.yml
Javagedes: Updated. They should match.

PR #371, opened 2022-12-15
Synchronize short commit id support

### LINKED ISSUE #236: Repo resolver support shorthand commits
**Describe the bug**
The edk2toolext\environment\repo_resolver.py doesn’t accept shorthand commit values.  To ensure the dependency file has a correct commit hash value, I suggest adding a check for a full commit hash value near line 215:

        else:
            if len(commit) != 40:
                logger.critical(
                    "Dependency {0} has a malformed commit hash.  Fail.".format(dep["Path"]))
                raise Exception(
                    "Dependency {0} has a malformed commit hash.  Fail.".format(dep["Path"]))
            if repo.head.commit == commit:

**To Reproduce**

**Expected behavior**

**Execution environment**
List the OS, python version, system environment variables, etc.

**Pip Packages**
If python related run `python -m pip list` and copy the output here

**Additional context**

Javagedes:
Short commit id's were partially supported in regards to the repo_resolver and git_dependency (ext_dep). This PR adds full support for both and adds tests to ensure they remain supported.

Resolves #236 

Signed-off-by: Joey Vagedes <joeyvagedes@microsoft.com>

### DIFF
--- a/docs/user/features/extdep.md
+++ b/docs/user/features/extdep.md
@@ -265,7 +265,7 @@ recommended to use this to ensure the contents are as expected.
 ## Git Type Schema Differences
 
 - source: url of git repo
-- version: commit hash to checkout
+- version: commit hash to checkout (short or long)
 
 ### Experimental Option: url_creds_var
 
--- a/docs/user/features/extdep.md
+++ b/docs/user/features/extdep.md
@@ -265,7 +265,9 @@ recommended to use this to ensure the contents are as expected.
 ## Git Type Schema Differences
 
 - source: url of git repo
-- version: commit hash to checkout (short or long)
+- version: commit hash to checkout (short\* or long)
+
+\* Short is considered the first 7 digits of a commit hash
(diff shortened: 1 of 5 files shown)


docs/user/features/extdep.md
```diff
@@ -265,7 +265,7 @@ recommended to use this to ensure the contents are as expected.
 ## Git Type Schema Differences
 
 - source: url of git repo
-- version: commit hash to checkout
+- version: commit hash to checkout (short or long)
```
makubacki, line 268: Though 7 digits is generally accepted as the default short hash length, I think it would be helpful to explicitly say something like "first 7 digits".

makubacki (approved):

docs/user/features/extdep.md
Javagedes: Added a note that a short commit is considered a the first 7 digits of a commit hash.

PR #375, opened 2022-12-20
Nuget Publishing deprecation: switch from LicenseUrl to License

Javagedes:
Using the LicenseUrl element within the .nuspec is now deprecated for nuget.exe. Users now have two ways to upload license information to a package:

1. Users must now pass a known SPDX identifier such that licenses.nuget.org/[license] is a valid license. This is now done by providing the `license` element as such: `<license type="expression">[License Type]</license>`.
2. Users may provide a custom license, but it must be uploaded with the package. It is also done via the `license` element, but by adding an additional `file` element to the `files` element as such: `<license type="file">license.txt (or license.md)</license>` and `<file src="[abs path to license file]" target="" />`

This PR changes `nuget_publishing.py` to meet these requirements by updating the "New" and "Pack" commands as follows:

## "New" Command

**Arguments**
* Adds: --CustomLicense (a flag) to specify the license is a file and will be provided in the pack command.
* Updates: --LicenseType command to --LicenseIdentifier, working exactly the same.

**Logic Change**
These commands both set the license field in the config yaml. --LicenseIdentifier set's it to a valid license type while --CustomLicense set's it to a string informing the user to specify the absolute path to the license in the "Pack" Command

## "Pack" Command

**Arguments**
* Adds: --CustomLicensePath to specify the absolute path to the license when packingthe nuget package

**Logic Change**
This command now verifies if the license is valid. If it is not, it will attempt to load it from the --CustomLicensePath argument. If that fails, it will raise an exception and fail. There is also logic change in _MakeNuspecXml() to update the license element according to if it is a file or expression. If it is a file, it adds a file element pointing to the license file to load.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -129,6 +129,7 @@
         "rtags",
         "Pdbs",
         "uncrustify",
-        "mkdocs"
+        "mkdocs",
+        "pygit"
     ]
 }
--- a/.github/workflows/doc-release.yml
+++ b/.github/workflows/doc-release.yml
@@ -1,0 +1,61 @@
+# This workflow builds and deploys documentation on a release.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+
+name: Doc Deployment
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    name: Build and Upload Site Artifact
+    
+    runs-on: ubuntu-latest
+
+    env:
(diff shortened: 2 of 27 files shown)


Javagedes:
The way I do test cleanup is currently breaking unit tests. Will fix.

makubacki:
It looks like this should roll a 0.21 release due to the LicenseUrl deprecation changes.

makubacki (approved):

cfernald (approved):

PR #377, opened 2022-12-29
Update edk2_git.py to use PyGit2 [REBASE&FF]

Javagedes:
Updates Edk2_git.py to use PyGit2 as backbone of class. Originally, the class was relied heavily on `RunCmd` to get much of the status information used by the `Repo` class. As a side effect of the implementation, these status indicators such as self.dirty are not updated in the class changes.

By using PyGit2, the git database (.git) is now parsed, which is quicker than using RunCmd. Due to this change, the status indicators are now dynamic and will update if the repo updates.

Some changes did ripple across the repo as PyGit2 holds references to certain files, so these references need to be freed if attempting to delete the repo.

Additionally, a small change was made in self_describing_environment so that it no longer uses PyGit2 directly, but uses edk2_git.py

### DIFF
--- a/edk2toolext/edk2_git.py
+++ b/edk2toolext/edk2_git.py
@@ -15,6 +15,10 @@ import os
 import logging
 from io import StringIO
 from edk2toollib.utility_functions import RunCmd
+import pygit2 as git
+from pathlib import Path
+from typing import Union  # Backwards compat for 3.9
+from edk2toollib.utility_functions import RemoveTree
 
 
 class ObjectDict(object):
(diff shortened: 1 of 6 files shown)


spbrogan (approved):
assuming your unit tests pass and coverage is high enough i think this code refactoring is too big for me to give it a good review.

edk2toolext/edk2_git.py
```diff
@@ -43,139 +47,170 @@ def set(self, key, value):
 
 
 class Repo(object):
```
makubacki, line 49: I find the existence of this abstraction within edk2-pytool-extensions confusing. To the casual reader, it appears to introduce redundant and narrowly scoped interfaces obscuring a more stable and broadly applicable interface directly provided by pygit2.

I understand that to some extent, it may provide minor convenience wrappers, but I don't think that justifies the additional layer of interface complexity alone.

Is it documented somewhere what value it provides over interacting with pygit2 directly?

Javagedes: @makubacki This will be abandoned in favor of GitPython and completely removing the edk2_git abstraction.

PR #378, opened 2022-12-29
Use reusable workflows to set variables used by multiple workflows in one place

Javagedes:
Previous PR (#365) was closed because it was based off my personal master branch and my changes got deleted. Reopening with with the PR based off of a branch.

Currently this is only used to set the python version and node version in a single place for all workflows, but can be extended.


spbrogan (approved):
looks good but again i'll rely on you as there is a lot of refactoring/structural changes being made

PR #379, opened 2022-12-29
platform build SDE failure description

Javagedes:
Adds additional information to the Runtime error produced by the SDE when external dependencies are out of date.


spbrogan (approved):

makubacki (commented):
While you're here, I'd consider replacing / elaborating upon "SDE". Most people probably don't know what that means, and it's commonly used for other topics like "software development environment".

PR #380, opened 2022-12-29
edk2_ci_setup: ensure repository with branch dependency is up to date

Javagedes:
Within edk2_ci_setup, if a repo dependency has a branch requirement, we only verify that the repo is on the correct branch, but not that the branch is up to date itself. This PR will perform a git pull (if update_ok / force is enabled) and update the branch to latest.


edk2toolext/environment/repo_resolver.py
```diff
@@ -286,6 +286,7 @@ def checkout(abs_file_system_path, dep, repo, update_ok=False, ignore_dep_state_
                 logger.info("We failed to checkout this branch, we'll try to fetch")
                 repo.fetch(branch=branch)
                 repo.checkout(branch=branch)
+            repo.pull()  # Pull to be top of tree
```
spbrogan, line 289: not sure why we need to pull here.  I worry this could create unintended consequences if repo is dirty.  Maybe we could pull with "--ff-only" but really still not sure why fetch and checkout isn't enough if we are using a fully defined repo path like origin/branch.

PR #387, opened 2023-01-05
Allow plugins to run on single module builds

Javagedes:
Allows plugins to run on single module builds by inherting from `IUefiSingleModuleBuildPlugin`

Examples:

```python
def MyPlugin(IUefiBuildPlugin):  # Runs only on a platform build
...

def MyPlugin(IUefiSingleModuleBuildPlugin):  # Runs only on a single module build
...

def MyPlugin(IUefiBuildPlugin, IUefiSingleModuleBuildPlugin):  # Runs on both platform and single module builds
...
```


spbrogan (changes requested):
As discussed, i don't like two independent classes which have same function name with same parameters.  We discussed other options, please follow through with that.

PR #393, opened 2023-01-11
Shell and build variable deletion

### LINKED ISSUE #132: Question: shell_environment:set_shell_var can i delete a variable
Can i use None or "" to delete a variable?

Javagedes:
Adds shell and build variable deletion functionality from within `ShellEnvironment` by accepting `None` as the variable value for functions `set_shell_var` and `set_build_var`.

Additionally fixes a spelling error `Overrideable ->  Overridable`. This spelling error was for an attribute of EnvEntry, making it a breaking change, however it should have minimal to no impact as it is used only by VarDict.

Closes #132

### DIFF
--- a/edk2toolext/environment/var_dict.py
+++ b/edk2toolext/environment/var_dict.py
@@ -30,7 +30,7 @@ class EnvEntry(object):
         """Inits an entry with the specified values."""
         self.Value = value
         self.Comment = comment
-        self.Overrideable = overridable
+        self.Overridable = overridable
 
     def PrintEntry(self, f=None):
         """Prints the value.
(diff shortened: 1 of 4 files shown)


edk2toolext/environment/var_dict.py
```diff
@@ -138,22 +164,29 @@ def GetValue(self, k, default=None):
...
-            (bool): If the variable was successfully stored or not
+            (bool): If the variable was successfully [stored|deleted] or not
         """
         key = k.upper()
         en = self.GetEntry(key)
         value = str(v)
-        self.Logger.debug("Trying to set key %s to value %s" % (k, v))
-        if (en is None):
+
+        if not v:
```
spbrogan, line 184: This is bad syntax.  What happens if V == 0.

spbrogan: see https://peps.python.org/pep-0008/#programming-recommendations for how comparing for none should be handled.

Javagedes:
Close in favor of #432

PR #417, opened 2023-01-25
Add CLI env information to the argument parser --help

Javagedes:
Adds a method to Edk2Invocable, AddParserEpilog(), that allows an invocable to customize the epilog of the argument parser when --help is used. In particular, Edk2PlatformBuild uses this method to append platform specified critical build env variables so that developers no longer need to dig through the code to see what build customizations are available.

Example output:
`stuart_build -c Platforms\QemuQ35Pkg\PlatformBuild.py --help`
`C:\src\mu_tiano_platforms\Platforms\QemuQ35Pkg> py PlatformBuild.py --help`

```cmd
options:
  -h, --help            show this help message and exit
  --SKIPBUILD, --skipbuild, --SkipBuild
                        Skip the build process
  --SKIPPREBUILD, --skipprebuild, --SkipPrebuild
                        Skip prebuild process
  --SKIPPOSTBUILD, --skippostbuild, --SkipPostBuild
                        Skip postbuild process
  --FLASHONLY, --flashonly, --FlashOnly
                        Flash rom after build.
  --FLASHROM, --flashrom, --FlashRom
                        Flash rom. Rom must be built previously.
  --UPDATECONF, --updateconf, --UpdateConf
                        Update Conf. Builders Conf files will be replaced with latest template files
  --CLEAN, --clean, --CLEAN
                        Clean. Remove all old build artifacts and intermediate files
  --CLEANONLY, --cleanonly, --CleanOnly
                        Clean Only. Do clean operation and don't build just exit.
  --OUTPUTCONFIG OUTPUTCONFIG, --outputconfig OUTPUTCONFIG, --OutputConfig OUTPUTCONFIG
                        Provide shell variables in a file
  -a BUILD_ARCH, --arch BUILD_ARCH
                        Optional - CSV of architecture to build. IA32,X64 will use IA32 for PEI and X64 for DXE and is the only valid option for this    
                        platform.
  --codeql              Optional - Produces CodeQL results from the build. See MU_BASECORE/.pytool/Plugin/CodeQL/Readme.md for more information.
  --build-config BUILD_CONFIG
                        Provide shell variables in a file
  --verbose, --VERBOSE, -v
                        verbose

CLI Env Variables:
  BLD_*_MEMORY_PROTECTION - If Memory protections are enabled.                       [TRUE]
  EMPTY_DRIVE             - Clears the Virtual Drive. We also allow for wrapping     [FALSE]
                            extremely long lines. See an example on this line. It
                            is very long!  It will also ignore new lines.
  QEMU_HEADLESS           - Run Qemu on Terminal Only.                               [FALSE]
  RUN_TESTS               - Runs tests specified by TEST_REGEX.                      [FALSE]
  SHUTDOWN_AFTER_RUN      - Shutdown after executing Startup.nsh.                    [FALSE]
  TARGET                  - The build target.                                        [DEBUG]
  TEST_REGEX              - Tests to Run.                                            [FALSE]
  TOOL_CHAIN_TAG          - The build toolchain.                                     [VS2019]

CLI Env Guide:
  <key>=<value>              - Set an env variable for the pre/post build process
  <key>                      - Set a non-valued env variable for the pre/post build process
  BLD_*_<key>=<value>        - Set a build flag for all build types
                               (key=value will get passed to build process)
  BLD_*_<key>                - Set a non-valued build flag for all build types
  BLD_<TARGET>_<key>=<value> - Set a build flag for build type of <target>
                               (key=value will get passed to build process for given build type)
  BLD_<TARGET>_<key>         - Set a non-valued build flag for a build type of <target>
```

The output is generated by implementing `SetPlatformDefaultEnv()` in the PlatformBuild.py which returns must return a list of `PlatformEnv` objects.

Here is an exmaple:

```python
def SetPlatformDefaultEnv(self) -> list[namedtuple]:
        """Returns a list of Env Variables that customize the build.

        Variables set here are printed to the terminal when executing
        stuart_build -c <Platform> --help.
        """
PlatformEnv = namedtuple('PlatformEnv', 'name description default')        
return [
            PlatformEnv("TOOL_CHAIN_TAG", "The build toolchain.", "VS2019"),
            PlatformEnv("TARGET", "The build target.", "DEBUG"),
            PlatformEnv("EMPTY_DRIVE", "Clears the Virtual Drive.", "FALSE"),
            PlatformEnv("RUN_TESTS", "Runs tests specified by TEST_REGEX.", "FALSE"),
            PlatformEnv("TEST_REGEX", "Tests to Run.", "FALSE"),
            PlatformEnv("QEMU_HEADLESS", "Run Qemu on Terminal Only.", "FALSE"),
            PlatformEnv("SHUTDOWN_AFTER_RUN", "Shutdown after executing Startup.nsh.", "FALSE"),
            PlatformEnv("BLD_*_MEMORY_PROTECTION", "If Memory protections are enabled.", "TRUE")
        ]

### DIFF
--- a/edk2toolext/environment/uefi_build.py
+++ b/edk2toolext/environment/uefi_build.py
@@ -28,15 +28,7 @@ from edk2toollib.utility_functions import RunCmd, RemoveTree
 from edk2toolext import edk2_logging
 from edk2toolext.environment.plugintypes.uefi_build_plugin import IUefiBuildPlugin
 import datetime
-
-
-class PlatformEnv():
-    """Container representing a Platform Environment Variable."""
-    def __init__(self, name, description, default):
-        """Inits a Platform Environment Variable."""
-        self.name = name
-        self.description = description
-        self.default = default
+from collections import namedtuple
 
 
 class UefiBuilder(object):
@@ -133,14 +125,6 @@ class UefiBuilder(object):
             self.SkipPreBuild = True
             self.SkipPostBuild = True
             self.FlashImage = False
-
-    def DeclareCriticalPlatformEnv(self) -> list[PlatformEnv]:
-        """Returns a list of Env Variables that customize the build.
-
-        Variables set here are printed to the terminal when executing
-        stuart_build -c <Platform> --help.
-        """
-        return []
 
     def Go(self, WorkSpace, PackagesPath, PInHelper, PInManager):
         """Core executable that performs all build steps."""
@@ -506,6 +490,10 @@ class UefiBuilder(object):
         # set environment variables for the build process
         os.environ["EFI_SOURCE"] = self.ws
 
+        # set critical platform Env Defaults if not set anywhere else in the build.
+        for env_var in self.SetPlatformDefaultEnv():
+            self.env.SetValue(env_var.name, env_var.default, "Default Critical Platform Env Value.")
+
         return 0
 
     def FlashRomImage(self):
@@ -558,6 +546,24 @@ class UefiBuilder(object):
             (int): 0 on success, 1 on failure
         """
         return 0
+
+    @classmethod
+    def SetPlatformDefaultEnv(self) -> list[namedtuple]:
+        """Sets platform default environment variables by returning them as a list.
+
+        Variables returned from this method are printed to the command line when
+        calling stuart_build with -h, --help. Variables added here should be
+        reserved only those that are commonly overwritten in the command line for
+        developers.
+
+        Variables returned from this function are the last variables to be set,
+        ensuring that default values are only added if none have been provided by
+        other means.
+
+        Returns:
+            (list[namedtuple]): List of named tuples containing name, description, default
+        """
+        return []
 
     @classmethod
     def PlatformBuildRom(self):
--- a/edk2toolext/environment/uefi_build.py
+++ b/edk2toolext/environment/uefi_build.py
@@ -30,6 +30,15 @@ from edk2toolext.environment.plugintypes.uefi_build_plugin import IUefiBuildPlug
 import datetime
 
 
+class PlatformEnv():
+    """Container representing a Platform Environment Variable."""
+    def __init__(self, name, description, default):
+        """Inits a Platform Environment Variable."""
+        self.name = name
+        self.description = description
+        self.default = default
+
+
 class UefiBuilder(object):
     """Object responsible for the full build process.
 
@@ -124,6 +133,14 @@ class UefiBuilder(object):
             self.SkipPreBuild = True
             self.SkipPostBuild = True
             self.FlashImage = False
+
+    def DeclareCriticalPlatformEnv(self) -> list[PlatformEnv]:
+        """Returns a list of Env Variables that customize the build.
+
+        Variables set here are printed to the terminal when executing
+        stuart_build -c <Platform> --help.
+        """
+        return []
 
     def Go(self, WorkSpace, PackagesPath, PInHelper, PInManager):
         """Core executable that performs all build steps."""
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by makubacki — APPROVED** [review]

File: edk2toolext/invocables/edk2_platform_build.py [hunk 1]
```diff
@@ -74,6 +74,25 @@ def RetrieveCommandLineOptions(self, args):
...
+        variables = self.PlatformBuilder.DeclareCriticalPlatformEnv()
+        if any(variables):
+            variables.sort(key=lambda v: v.name)
+
+            max_name_len = max(len(var.name) for var in variables)
+            max_desc_len = max(len(var.description) for var in variables)
+
+            custom_epilog += "CLI Env Variables:"
+            for v in variables:
+                custom_epilog += f"\n  {v.name:<{max_name_len}} - {v.description:<{max_desc_len}}  [{v.default}]"
```
spbrogan, line 91: i like the right padding trick.

edk2toolext/environment/uefi_build.py
```diff
@@ -30,6 +30,15 @@
 import datetime
 
 
+class PlatformEnv():
+    """Container representing a Platform Environment Variable."""
+    def __init__(self, name, description, default):
```
spbrogan, line 35: please add typing

edk2toolext/environment/uefi_build.py
```diff
@@ -125,6 +134,14 @@ def RetrievePlatformCommandLineOptions(self, args):
             self.SkipPostBuild = True
             self.FlashImage = False
 
+    def DeclareCriticalPlatformEnv(self) -> list[PlatformEnv]:
```
spbrogan, line 137: This function should be moved down near https://github.com/tianocore/edk2-pytool-extensions/blob/b83d8c5fbf7398a5dd1a1cacdf00b1afc326375b/edk2toolext/environment/uefi_build.py#L516

And Probably start with "Platform" and have classmethod decorator (although i don't know if this does anything beyond the default anymore).

spbrogan:
I would expect documentation update to match.  For this feature to really deliver we need to make sure all platform developers are aware of the new capability and move their "defaults" into the declare list.

spbrogan:
I would expect documentation update to match.  For this feature to really deliver we need to make sure all platform developers are aware of the new capability and move their "defaults" into the declare list.

spbrogan (commented):
I would expect documentation update to match.  For this feature to really deliver we need to make sure all platform developers are aware of the new capability and move their "defaults" into the declare list.

edk2toolext/environment/uefi_build.py
Javagedes: PlatformEnv class removed in preference of a namedtuple

Javagedes:
> Please also test out code paths where users don't supply -C file and/or have an invalid file. Also probably need to test out the other split of SettingsInterface.

The logic of not supplying `-C <path>` or incorrectly providing a path remains unchanged as this is managed by the first argparser. The second argparser, which is what provides the platform env information added in this PR is only instantiated once the first argparser has successfully found and verified the platform module file.

AddParserEpilog() has a default epilog, so all other settings file remains unchanged unless we decide we want them to provide a different epilog. I went ahead and tested each of our invocables and they all work as intended.

I also verified that running the platformbuild.py directly (if supported) works as intended (i.e. `py PlatformBuild.py -h`, `py PlaformBuild.py --setup -h`, etc.)

edk2toolext/environment/uefi_build.py
Javagedes: Renamed to `SetPlatformDefaultEnv()` as it seems like an intuitive and accurate name. The variables returned from `SetPlatformDefaultEnv()` are now also the last variables to be set in `SetEnv()`.

edk2toolext/invocables/edk2_platform_build.py
Javagedes: Additionally, just added line wraps for long descriptions.

spbrogan (approved):

PR #424, opened 2023-01-31
nuget_publishing: Make `license` entry in config file optional

makubacki:
Previously the NuGet config file required a dummy license value to
be present if `--CustomLicensePath` was given. This change removes
the dummy value being required.

Co-authored-by: Joey Vagedes <joeyvagedes@microsoft.com>


makubacki:
As discussed offline, @Javagedes is going to make some additional changes

Javagedes (approved):

PR #429, opened 2023-02-02
nuget_publishing: Add repository metadata support

makubacki:
NuGet packages can associate a package with source repositories using
the `repository` metadata element.

The following for optional attributes are allowed per the corresponding
nuspec version:

- `type` (4.0+)
- `url` (4.0+)
- `branch` (4.6+)
- `commit` (4.6+)

See the following nuspec reference for more info:
https://learn.microsoft.com/en-us/nuget/reference/nuspec#repository

The GitHub NuGet registry (among others) can use this information to
associate a given package with the specified source content.

See the following GitHub doc for more info:
https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry

This change adds support for the `repository` element in both the
`New` (set in config file) and `Pack` (change at cmd line) commands
so it can be used when publishing NuGet packages.

### DIFF
--- a/edk2toolext/nuget_publishing.py
+++ b/edk2toolext/nuget_publishing.py
@@ -48,6 +48,7 @@ class NugetSupport(object):
         <license></license>
         <releaseNotes></releaseNotes>
         <projectUrl></projectUrl>
+        <repository />
         <copyright></copyright>
         <tags></tags>
     </metadata>
@@ -113,7 +114,9 @@ class NugetSupport(object):
         with open(self.Config, "r") as c:
             self.ConfigData = yaml.safe_load(c)
 
-    def SetBasicData(self, authors, license, project, description, server, copyright):
+    def SetBasicData(self, authors, license, project, description, server, copyright,
+                     repositoryType=None, repositoryUrl=None, repositoryBranch=None,
+                     repositoryCommit=None):
         """Set basic data in the config data."""
         self.ConfigData["author_string"] = authors
         if license:
@@ -121,7 +124,14 @@ class NugetSupport(object):
         self.ConfigData["project_url"] = project
         self.ConfigData["description_string"] = description
         self.ConfigData["server_url"] = server
-
+        if repositoryType:
+            self.ConfigData["repository_type"] = repositoryType
+        if repositoryUrl:
+            self.ConfigData["repository_url"] = repositoryUrl
+        if repositoryBranch:
+            self.ConfigData["repository_branch"] = repositoryBranch
+        if repositoryCommit:
+            self.ConfigData["repository_commit"] = repositoryCommit
         if not copyright:
             copyright = "Copyright %d" % datetime.date.today().year
         self.ConfigData["copyright_string"] = copyright
@@ -165,6 +175,21 @@ class NugetSupport(object):
         """Update tags in the config data."""
         self.ConfigData["tags_string"] = " ".join(tags)
         self.ConfigChanged = True
+
+    def UpdateRepositoryInfo(self, type=None, url=None, branch=None, commit=None):
+        """Update repository information."""
+        if type:
+            self.ConfigData["repository_type"] = type
+            self.ConfigChanged = True
+        if url:
+            self.ConfigData["repository_url"] = url
+            self.ConfigChanged = True
+        if branch:
+            self.ConfigData["repository_branch"] = branch
+            self.ConfigChanged = True
+        if commit:
+            self.ConfigData["repository_commit"] = commit
+            self.ConfigChanged = True
 
     def Print(self):
         """Print info about the Nuget Object."""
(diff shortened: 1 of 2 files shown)


makubacki:
I know there's fancier ways to organize the changes, but this is meant to align with the code structure currently in place. Not that I see significant functional value in refactoring the file given its current length.

edk2toolext/nuget_publishing.py
```diff
@@ -217,6 +242,17 @@ def _MakeNuspecXml(self, ContentDir, ReleaseNotesText=None):
         meta.find("version").text = self.NewVersion
         meta.find("authors").text = self.ConfigData["author_string"]
         meta.find("projectUrl").text = self.ConfigData["project_url"]
+        repository_item_present = bool([k for k in self.ConfigData.keys() if "repository" in k.lower()])
+        if repository_item_present:
```
Javagedes, line 246: Repository should be removed if no items are present. `meta.remove()`

edk2toolext/nuget_publishing.py
```diff
@@ -368,6 +404,14 @@ def GatherArguments():
                             required=True)
         parser.add_argument('--Author', dest="Author", help="<Required> Author string for publishing", required=True)
         parser.add_argument("--ProjectUrl", dest="Project", help="<Required> Project Url", required=True)
+        parser.add_argument("--RepositoryType", dest="RepositoryType", help="<Optional> Repository Type",
```
Javagedes, line 407: I would add these to an argument group. `g = parser.add_argument_group(title = "", description="")`. Then `g.add_argument()`. It doesn't do anything extra logically but will help the user out a bit when doing `-h, --help`

edk2toolext/nuget_publishing.py
```diff
@@ -395,6 +439,14 @@ def GatherArguments():
         parser.add_argument('--CustomLicensePath', dest="CustomLicensePath", default=None,
                             help="<Optional> If CustomLicense set in `new` phase, provide absolute path of License \
                             File to pack. Does not override existing valid license.")
+        parser.add_argument("--RepositoryType", dest="RepositoryType", help="<Optional> Repository Type",
```
Javagedes, line 442: Argument group as mentioned above.

Javagedes (commented):
I'm not a huge fan of needing to pass all four of these repo details around everywhere. You can change it however you like to solve this, however one thought I had is to create a dict if any of those values are set with `if any([args.RepositoryType, args.RepositoryUrl, args.RepositoryBranch, args.RepositoryCommit]):`

For `new` you can pass the dict to `nu.SetBasicData()` and set the repo details in the config, which yaml should have no problem dumping and loading a dict, even when a value = None.

For `pack` \ `packandpush` you can now check if that dict is None or not, and if its not, call `updaterepositoryinfo()` on the dict and set it directly.

Logic wouldn't need to change too much in _makenuspecxml other then indexing `if <type/url/branch/commit> in self.ConfigData["repository"]`

makubacki:
> I'm not a huge fan of needing to pass all four of these repo details around everywhere.

This is why I prefaced the change with the comment:

> I know there's fancier ways to organize the changes, but this is meant to align with the code structure currently in place. Not that I see significant functional value in refactoring the file given its current length.

If I'm going to make any adjustments for this change, I'd rather just update the rest of the file since it's not written for extensibility as-is and I don't think micro data organization adjustments make much of a difference in the larger existing file structure.

edk2toolext/nuget_publishing.py
makubacki: Good point. I'll add that.

Javagedes:
> > I'm not a huge fan of needing to pass all four of these repo details around everywhere.
> 
> This is why I prefaced the change with the comment:
> 
> > I know there's fancier ways to organize the changes, but this is meant to align with the code structure currently in place. Not that I see significant functional value in refactoring the file given its current length.
> 
> If I'm going to make any adjustments for this change, I'd rather just update the rest of the file since it's not written for extensibility as-is and I don't think micro data organization adjustments make much of a difference in the larger existing file structure.

While I agree it needs a rewrite, I would argue it is not worth the investment at the moment. I'll add it to my backlong.

edk2toolext/nuget_publishing.py
makubacki: Done in PR commit https://github.com/tianocore/edk2-pytool-extensions/pull/429/commits/2682256cc76730f18db5526ac045b621f119f4db

edk2toolext/nuget_publishing.py
makubacki: Done in PR commit https://github.com/tianocore/edk2-pytool-extensions/pull/429/commits/2682256cc76730f18db5526ac045b621f119f4db

edk2toolext/nuget_publishing.py
makubacki: Done in PR commit https://github.com/tianocore/edk2-pytool-extensions/pull/429/commits/2682256cc76730f18db5526ac045b621f119f4db

makubacki:
> While I agree it needs a rewrite, I would argue it is not worth the investment at the moment. I'll add it to my backlong.
> 
> I'm fine without adding the consolidation of repository. data.

I made the other updates requested. I'm also happy to refactor the file as well. Just let me know what you'd finally like to do.

Javagedes (approved):

PR #432, opened 2023-02-06
shell_environment: set_shell_var() exception

Javagedes:
Add an exception to `set_shell_var()` when var_data is None as it is not supported. It will throw an error when it attempts to set os.environ[var_name] = None. Adds this information to the pydocs for the function.

### DIFF
--- a/edk2toolext/environment/shell_environment.py
+++ b/edk2toolext/environment/shell_environment.py
@@ -319,15 +319,21 @@ class ShellEnvironment(metaclass=Singleton):
         return self.active_buildvars.GetValue(var_name)
 
     def set_build_var(self, var_name, var_data):
-        """Sets the build var.
+        """Sets the variable as internal build variable.
+
+        !!! note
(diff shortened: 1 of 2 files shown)


edk2toolext/environment/shell_environment.py
```diff
@@ -344,15 +349,18 @@ def get_shell_var(self, var_name):
...
         Args:
             var_name (str): variable to set the value for
             var_data (obj): data to set
 
-        The variable is set both in the `VarDict` and in the OS
+        Raises:
+            (ValueError): var_data is None
         """
         # Check for the "special" shell vars.
+        if var_data is None:
```
makubacki, line 362: It would be nice if the exception path was covered by a simple unit test.

makubacki (approved):

edk2toolext/environment/shell_environment.py
Javagedes: Test Added. Thanks!

PR #443, opened 2023-02-13
edk2_logging Updates [Rebase & FF]

makubacki:
Includes 3 commits.

---

## Commit 1 - Add initial set of logging unit tests

No unit tests previously existed specifically for `edk2_logging`.

Adds tests to verify edk2_logging behavior.

## Commit 2 - Strip whitespace from errors

Currently whitespace is left in source and error substrings extracted
from the full line string from build output. This makes whitespace
around the strings that print unpredictable and impacts readability.

This change removes the extra whitespace and updates unit tests to
show how output is printed before and after the change.

Example Input:

`<source_file>error A1: error 1 details`

Example Output Before (Old):

`Compiler https://github.com/tianocore/edk2-pytool-extensions/pull/1 from <source_file>  error 1 details`

Example Output After (New):

`Compiler https://github.com/tianocore/edk2-pytool-extensions/pull/1 from <source_file> error 1 details`

## Commit 3 - Add GCC compilation error info to log

Adds information about GCC compilation errors (file and issue) to the
log similar to other messages printed so the developer can get a
better idea what went wrong at a high-level.

In this case, a compiler error number is not printed, so have the
pre-existing string dump print `"#error"`.


edk2toolext/edk2_logging.py
```diff
@@ -201,20 +201,28 @@ def scan_compiler_output(output_stream):
...
+    # than offering per rule flexibility to parse tokens via regex
+    gcc_error_exp = re.compile(r":\d+:\d+: (error):")
     edk2_error_exp = re.compile(r"error F(\d+):")
     build_py_error_exp = re.compile(r"error (\d+)E:")
     linker_error_exp = re.compile(r"error LNK(\d+):")
     warning_exp = re.compile(r"warning [A-Z]?(\d+):")
     for raw_line in output_stream.readlines():
         line = raw_line.strip("\n").strip()
         match = error_exp.search(line)
+        if match is not None:
```
Javagedes, line 222: Any thoughts on adding a break after we match any of the below if statements? I presume once we find one, we can move to the next line instead of checking the same line for all other error types.

Javagedes (approved):

edk2toolext/edk2_logging.py
makubacki: I'd need a bit more time to confirm that. Because the regex matches fragments of an overall line string it is possible (though I think unlikely) that more than one match occurs on a line.

I would prefer to hold off on that to prevent potentially breaking an obscure case until more investigation is done.

.cspell.json
```diff
@@ -9,7 +9,8 @@
     ],
     "allowCompoundWords": true,
     "ignorePaths": [
-        "*.exe"
+        "*.exe",
+        "edk2toolext/tests/test_edk2_logging.py"
```
Javagedes, line 13: Please change this to all tests. I'm not particularly worried about spelling in tests and I'd like to keep it consistent.

PR #449, opened 2023-02-21
edk2_logging: filter secrets [REBASE&FF]

Javagedes:
This commit updates the edk2_logging filter to substitute secrets with `*****` if found.


edk2toolext/edk2_logging.py
```diff
@@ -255,6 +255,13 @@ def __init__(self):
         self._verbose = False
         self._currentSection = "root"
 
+        secrets_regex_strings = [
+            r"[A-Za-z0-9]{46}",  # Nuget API Key
+            r"github_pat_[^ ]*"  # Github PAT
```
makubacki, line 260: The GitHub PATs I've seen of the format `ghp_<value>`.

Like what is described here: [Authentication token format updates are generally available](https://github.blog/changelog/2021-03-31-authentication-token-format-updates-are-generally-available/).

The prefix changes between token types. I think we should cover all defined types.

This won't cover the previous token format. I think that's fine given its ambiguity (that led GitHub to this change) and the fact the new format has been in effect for nearly two years now.

makubacki, line 260: I think it would be better to go with what GitHub has strictly allowed as PAT characters (`[A-Za-z0-9_]`).

For example, right now `<PAT>,<other data>` would include the text after the comma though the comma would end the PAT (just as much as a space) per GitHub's allowed character set.

makubacki, line 260: Might also be worth considering using `+` instead of `*` since a prefix without any matched characters doesn't need to be hidden.

Could also use `{<min>,}` if the min token length ever used is documented somewhere. But, I think `+` would help differentiate a little better between a real need to hide the value and something that might unnecessarily impact debuggability.

Javagedes: Thanks for sending me that documentation! I guess what I found was old. I've Updated to support gh[pousr]_.

Javagedes: Going with + over * and {<min>, 255} as I cant find anything about a min value, only that the max is 255. (I didin't look hard, and I think + is better anyway.)

Javagedes: The documentation you sent was great. I've updated to match and added a few tests.

makubacki (approved):

PR #465, opened 2023-03-07
Fixed setup --force when submodules don't exist

bkarstens:
If `stuart_setup` was called with the `--force` option when the submodules aren't initialized (either new clone or they were deinit'd), the setup would fail. This fixes that problem.

PR #467, opened 2023-03-08
Add edk2_setup tests

Javagedes:
Adds additional tests for handling a --FORCE flag when submodules are not initialized. Additonally adds testing when using windows style path separators (\\)


edk2toolext/environment/repo_resolver.py
```diff
@@ -418,24 +418,27 @@ def clean(abs_file_system_path, ignore_files=[]):
         repo.git.clean(*params)
 
 
-def submodule_clean(abs_file_system_path, submodule, ignore_files=None):
```
Javagedes, line 421: unecessarily added in #465

edk2toolext/environment/repo_resolver.py
```diff
@@ -418,24 +418,27 @@ def clean(abs_file_system_path, ignore_files=[]):
...
 
     Raises:
         (GitCommandError): The command is invalid
         (InvalidGitRepositoryError): The repo is in an invalid format
         (NoSuchPathError): The path does not exist
         (ValueError): submodule's path was invalid
     """
-    submodule_path = Path(submodule.path).as_posix()
     with Repo(abs_file_system_path) as repo:
-        if repo.submodule(submodule_path).module_exists():
```
Javagedes, line 437: `repo.submodule()` expects the submodule name, not path. For MU repos, the name and path are the same, but a submodule name does not have to be the relative path to the submodule, so we will find the submodule to update by comparing paths.

PR #473, opened 2023-03-20
dependabot.yml: Add GitHub actions

Javagedes:
Adds github-actions to the list of dependencies dependabot checks for.


makubacki (approved):
I suggest changing the PR title to "dependabot.yml: Add GitHub actions" or something similar to help call out the change better in history.

Issue #488, opened 2023-04-03
[Feature]: A utility to audit a systems dbx


Flickdm:
### What does the feature solve?

This feature would allow a user to audit the dbx contents on their system to determine what revocations exist on their system.

### Describe the solution

A tool grabs the dbx file from a system, and compares it against the public revocations found on UEFI.org from the csv file.  

### Have you considered any alternatives?

There are none.

Javagedes:
Closed via #489

PR #489, opened 2023-04-03
Add audit dbx tool

Flickdm:
linking issue #488 

This provides a tool and examples to get audit the dbx on a given system


apop5 (approved):

requirements.txt
```diff
@@ -7,4 +7,7 @@ pefile == 2023.2.7
 semantic_version == 2.10.0
 GitPython == 3.1.31
 pyyaml == 6.0.0
-cryptography == 40.0.1
\ No newline at end of file
+cryptography == 40.0.1
+openpyxl == 3.1.2
```
Javagedes, line 11: These need to be added to the setup.py file, install_requires. If this tool only works on Windows, I would go ahead and just add the windows requirement to openpyxl and XlsxWriter too, since they would only ever be used with the tool.

edk2toolext/windows/secureboot/secureboot_audit.py
Javagedes, line 1: You need to add the script entry point to the setup.py file's entry_points section.

docs/user/tools/using_secureboot_audit.md
```diff
@@ -0,0 +1,116 @@
...
+Limitations:
+
+    1. Currently only supports windows
+    2. The audit is only as good as the uefi.org csv file
+
+This tool has three primary stages:
+
+1. Get the dbx from UEFI
+2. Convert the [UEFI.org csv (xlsx)](https://uefi.org/revocationlistfile) of revocations to json format for comparison
+3. Parse the dbx file and compare it against the uefi.org revocation sheet
```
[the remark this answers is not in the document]
Flickdm: No I forgot to include the header for stage 3

edk2toolext/tests/test_secureboot_audit.py
```diff
@@ -0,0 +1,133 @@
...
+    generate_dbx_report,
+    filter_revocation_list_by_arch,
+    convert_uefi_org_revocation_file_to_dict,
+    write_xlsx_file,
+    write_json_file,
+)
+
+
+# Setup the test directory path
+TEST_DATA_PARENT_DIRECTORY = os.path.join(
```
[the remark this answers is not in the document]
Flickdm: So orignally the weren't meant for testing but they have become that. I'll go ahead and move them

edk2toolext/windows/secureboot/examples/dbx.bin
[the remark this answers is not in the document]
Flickdm: This is an efi signature list of hashes of efi binaries. It was meant as an example of a dbx.bin file which is exactly what the script would pull off a given system. I can serialize it, but it felt like it fell into the same category as the other files in this folder. What would you prefer?

edk2toolext/tests/test_secureboot_audit.py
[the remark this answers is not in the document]
Javagedes: Most people are going to access your tool by pip installing edk2-pytool-examples and using the command via the CLI. They won't be able to access the these files as is. If they are going to actually review the readme on how to use the tool, it will be via the documentation we automatically generate (See [example](https://www.tianocore.org/edk2-pytool-extensions/tools/using_image_validation_tool/)). Due to this, if you want to show an easily accessible example, I would take a screenshot, lower the quality of the image, and put it in your readme.

docs/user/tools/using_secureboot_audit.md
```diff
@@ -0,0 +1,116 @@
+# SecureBootReport.py
```
Javagedes, line 1: This first header is what shows as the tool name in the auto generated docs. This should be changed to Secureboot Report Tool (or something to that effect)

edk2toolext/tests/test_secureboot_audit.py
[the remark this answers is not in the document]
Flickdm: How do I do that? Only the get dbx command runs on windows, but the rest of it (such as converting the uefi.org xlsx file to json is platform agnostic)

docs/user/tools/using_secureboot_audit.md
```diff
@@ -0,0 +1,123 @@
...
+## Stage 2 (Convert the UEFI.org Excel Sheet)
+
+    ```bash
+    python SecureBootReport.py convert_file dbx_info_2020_2023_uefiorg_v3.xlsx
+    INFO:root:Wrote report to .\SecureBootFiles\uefi_org_revocations.json
+    ```
+
+Flags:
+
+1. `--output` [not required] - allows for redirection of the output to a new path and name
```
Javagedes, line 36: nit: not indented

edk2toolext/tests/test_secureboot_audit.py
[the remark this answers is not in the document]
Javagedes: Ah thats my misunderstanding then. If this works with more then just windows, then definitely run it on both operating systems. If there are any tests you would like to only run on windows (or only linux), you can use `@pytest.mark.skipif(sys.platform == "<something>")`

https://docs.pytest.org/en/6.2.x/skipping.html#id1

requirements.txt
```diff
@@ -7,4 +7,7 @@ pefile == 2023.2.7
...
+XlsxWriter == 3.0.9
+pywin32 == 305; sys_platform == 'win32'
```
Javagedes, line 13: nit: use 306 here. dependabot will just immediately yell at me. 305 in the setup.py is perfectly fine however as you are supposed to put the minimum viable version in the setup.py so that there is less chance of conflicts when people install multiple modules.

setup.py
```diff
@@ -77,15 +77,19 @@ def run(self): # noqa
                             'firmware_policy_tool=edk2toolext.windows.policy.firmware_policy_tool:main',
                             'edk2_capsule_tool=edk2toolext.capsule.capsule_tool:main',
                             'versioninfo_tool=edk2toolext.versioninfo.versioninfo_tool:main',
-                            'validate_image_tool=edk2toolext.image_validation:main']
+                            'validate_image_tool=edk2toolext.image_validation:main',
+                            'secureboot_audit=edk2toolext.windows.secureboot.secureboot_audit:main']
```
Javagedes, line 81: nit: add a comma at the end of this list. This is so that when we add a new tool, it shows we are only editing one line instead of two like what we see above (because I failed to add a comma to validate_image_tool when I added that 🥇 )

edk2toolext/windows/secureboot/secureboot_audit.py
```diff
@@ -0,0 +1,579 @@
...
+    "F22246C9E65BB4543CB892B6507C062196A788E5A190D2D5A877B94DFDFB2935",  # "Debian Certificate",
+]
+
+
+###################################################################################################
+# Helper functions
+###################################################################################################
+
+
+def write_json_file(report, output_file):
```
Javagedes, line 58: Sorry doug, I missed this the first time. Please also add type hinting to the function arguments :(

PR #490, opened 2023-04-04
repo_resolver.py: `repo_details()` to use first remote if origin does not exist

Javagedes:
In some scenarios, the origin branch does not exist; due to this hardcoding the URL to be the origin URL can fail. In this scenario, we default to the first existing remote's URL.

### DIFF
--- a/edk2toolext/environment/repo_resolver.py
+++ b/edk2toolext/environment/repo_resolver.py
@@ -238,7 +238,7 @@ def repo_details(abs_file_system_path):
             details["Dirty"] = repo.is_dirty(untracked_files=True)
             details["Initialized"] = True
             details["Url"] = repo.remotes.origin.url or repo.remotes[0].url
-            details["Url"] = repo.remotes.origin.url
+
     except (InvalidGitRepositoryError, NoSuchPathError):
         pass
     return details
--- a/edk2toolext/environment/repo_resolver.py
+++ b/edk2toolext/environment/repo_resolver.py
@@ -237,6 +237,7 @@ def repo_details(abs_file_system_path):
             details["Bare"] = repo.bare
             details["Dirty"] = repo.is_dirty(untracked_files=True)
             details["Initialized"] = True
+            details["Url"] = repo.remotes.origin.url or repo.remotes[0].url
             details["Url"] = repo.remotes.origin.url
     except (InvalidGitRepositoryError, NoSuchPathError):
         pass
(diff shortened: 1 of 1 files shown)


apop5 (approved):

makubacki:
I think this is okay - it already handles a relatively uncommon scenario. However, it might be able to find a better match in situations where the user has a branch checked out tracking an upstream remote. In that case, the remote tracked by the branch is what I'd expect to be used. Not sure if that's worth adding right now or not.

edk2toolext/environment/repo_resolver.py
```diff
@@ -245,7 +245,7 @@ def repo_details(abs_file_system_path):
             details["Bare"] = repo.bare
             details["Dirty"] = repo.is_dirty(untracked_files=True)
             details["Initialized"] = True
-            details["Url"] = repo.remotes.origin.url
+            details["Url"] = repo.remotes.origin.url if "origin" in repo.remotes else repo.remotes[0].url
```
makubacki, line 248: Should have added the other comment here so it wouldn't accidentally be missed. Adding a reference to it https://github.com/tianocore/edk2-pytool-extensions/pull/490#issuecomment-1498450550.

makubacki (approved):

edk2toolext/environment/repo_resolver.py
Javagedes: @makubacki This is possible with something along the lines of:

```python
branch_name = repo.head.ref.name
remote_name = repo.branches[branch_name].tracking_branch().remote_name
remote_url = repo.remotes[remote_name].url
```

However there are a few pieces of error handling that would also need to be added:
1.  repo.head.ref.name will raise an exception if you're not on a branch
2. tracking_branch() can return None, which will raise an exception when you call remote_name on it

If you are okay with the additional complexity instead of this one liner, I can add it. I'm not 100% sold on it being necessary enough, but I can definitely do it if you think it's important enough. 

I would update the logic to return the url in the following order:
1. The branch's remote if on a branch and if it has a tracking remote
2. the origin url
3. the first url in the remotes list

makubacki: I'd personally prefer for it to be added. As a user, it matches what I'd intuitively expect. There have been cases where I rename my remotes in some workspaces to more clearly describe two alternative upstreams (like "secure" and "non_secure") so I reduce likelihood of push mistakes and there is not an "origin".

The complexity looks like normal API interaction overhead, so I'm not too concerned about it.

Your priority list looks good to me.

Ultimately, go with what you're comfortable supporting. If you drop checking the tracking remote, it'd be a good idea to document the behavior somewhere, like the function description, for example.

Javagedes: @makubacki I've made the appropriate changes and checked the validity under these circumstances:

1. A normal branch checked out from origin returns origin
2. a remote branch checked out returns the remote 
3. A branch that has had it's upstream set to a remote returns the remote

Are there any situations I've missed?

Javagedes: Good call It's been updated such that if it does not meet one of the three criteria, it returns None for the url. Merging.

PR #510, opened 2023-04-28
Resolve CI failures [REBASE&FF]

Javagedes:
Resolves two CI failures in edk2-pytool-extensions:

## Commit 1
Integration tests were failing due to the addition of the command line
build variable CODE_COVERAGE being set to true by default. This adds the
flag and sets it to false in integration tests as this execution is a
build plugin and outside the scope of edk2-pytool-extensions.

## Commit 2

With recent changes to pip, using a method from the package during the
setup of the package is no longer supported. Due to this, it was decided
to move when we download NuGet to when we first attempt to use it. This
has multiple benifits as we now no longer download NuGet, even if the
user is supplying their own version of NuGet, and our download NuGet
functionality remains inside the package and easily testable, rather
then an extra function inside of setup.py

Additionally we no longer use the deprecated module pkg_resources to
identify the binary folder inside the package, instead using
importlib's resources module to do this.

### DIFF
--- a/edk2toolext/bin/nuget.py
+++ b/edk2toolext/bin/nuget.py
@@ -10,6 +10,8 @@ import os
 import urllib.error
 import urllib.request
 import logging
+from importlib import resources
+from pathlib import Path
 
 # Update this when you want a new version of NuGet
 VERSION = "6.4.0"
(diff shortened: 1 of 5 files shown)


azure-pipelines/templates/basic-setup-steps.yml
```diff
@@ -21,5 +21,5 @@ steps:
 - script: pip install --upgrade -r requirements.txt
   displayName: 'Install requirements'
 
-- script: pip install -e .
+- script: pip install .
   displayName: 'Install from Source'
```
kuqin12, line 25: Is this comment still right if we drop `-e`?

kuqin12 (approved):

azure-pipelines/templates/basic-setup-steps.yml
Javagedes: We no longer are dropping the `-e .`. I identified a deeper issue with the recent pip updates that I was able to resolve :)

makubacki (approved):

Issue #502, opened 2023-04-20
[Bug]: Secret filter is obfuscating non-secrets in logs


makubacki:
### Describe the Bug

The NuGet API key filter is very generic and causing a number of strings to be replaced in build logs. Example:

d:\src\Build\QemuQ35Pkg\DEBUG_VS2022\FV\Ffs\EAEEDEF9-ABE7-4B95-82B0-*******\EAEEDEF9-ABE7-4B95-82B0-5A534C899B46.map

Here's another example of a filename filtered:

\*******BootPath.c(54):

### What Python version are you using?

Python 3.11

### Reproduction steps

1. Construct a file name that meets the character and length requirements
2. Check the build log

### Expected behavior

Try to reduce matches against non-secret strings. Some false positives are okay (better to side on safety). The request is to try to reduce the number of non-secret strings masked out.

PR #513, opened 2023-05-01
edk2_logging.py: increase specificity of NuGet API key

Javagedes:
The False positive rate of API keys being filtered out was a detriment to developer productivity per #502. This commit reduces the API key to be only lowercase a-z as NuGet API keys generated from NuGet.org are generated as all lowercase. This should reduce the amount of false positives noted, specifically false positives due to the long folder names in the build output folder.


makubacki (approved):
Most of the occurrences are Pascal named file paths, so I think this will help.

PR #516, opened 2023-05-01
setup.py: move cryptography requirement to extra_require's openssl

Javagedes:
The cryptography requirement was incorrectly placed in the normal requirements for edk2-pytool-extensions, however it is only actually required if using openssl functionality, therefore it should only be installed when using `pip install edk2-pytool-extensions[openssl]`


Javagedes:
Dropping this in favor of #523

PR #523, opened 2023-05-05
Transition to pyproject.toml [REBASE&FF]

Javagedes:
Adds three commits to transition project management, builds, and dev to pyproject.toml file.

Commit 1 renames `integration_tests/` to `tests.integration/` and moves/renames `edk2toolext/tests` to `tests.unit/`
Commit 2 is the core commit to view. It has the actual changes from setup.py to pyproject.toml.
Commit 3 Updates azure pipelines / github actions to work correctly.


pyproject.toml
```diff
@@ -0,0 +1,105 @@
+[build-system]
+requires = ["setuptools", "setuptools_scm[toml]"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "edk2-pytool-extensions"
+authors = [
```
makubacki, line 7: I think this should be a generic account (like `edk2_pytools@microsoft.com`). We have a lot of stale personal contact info in various places today.

makubacki (approved):

pyproject.toml
Javagedes: I have updated the email and made the request. I will hold off on merging this until the account is created.

PR #541, opened 2023-05-10
Revert skip verification

Javagedes:
The ability to skip external dependency verification was added with the anticipation of additional developer workflows, however those workflows have evolved over time and the ability to skip external dependency verification is no longer necessary. As it is no longer necessary, it only adds additional complexity to the codebase and confusion to developers attempting to use it. Removing this functionality is the best way forward, but can always be reconsidered or reworked in the future if necessary.


cfernald (approved):

cfernald:
Consider providing context in the PR description

PR #547, opened 2023-05-12
edk2_logging: add gcc fatal error regex

Javagedes:
Adds a regex that detects `fatal error` in the compiler output and logs the line as an error.

Example:
`BrotliCompress.c:20:10: fatal error: ./brotli/c/common/constants.h: No such file or directory`


makubacki:
@Javagedes, we discussed some more fundamental changes to the regular expressions. I assume you're going to complete this PR as-is and do that separately? I think it makes sense to go ahead and get this into a smaller release before that.

makubacki (approved):

Javagedes:
> @Javagedes, we discussed some more fundamental changes to the regular expressions. I assume you're going to complete this PR as-is and do that separately? I think it makes sense to go ahead and get this into a smaller release before that.

Yes I'll merge this in a small release then work on a rework later.

PR #575, opened 2023-05-26
edk2_invocable: verbose flag may not be honored

Javagedes:
A platform settings GetLoggingLevel() currently takes president over the `--Verbose` flag, but logically, a command flag should override the default behavior and set the logging level regardless of platform settings.


edk2toolext/edk2_invocable.py
```diff
@@ -260,14 +260,17 @@ def GetLoggingLevel(self, loggerType):
         Returns:
             (logging.Level): logging level
         """
+        if loggerType == "con" and self.Verbose:
```
VivianNK, line 263: Why only "con" (console - on screen print)? Is "txt" always verbose?

VivianNK (commented):
Overriding defaults when using --verbose makes sense to me. I think all debug messages should be accessible without getting into platform settings. 
Without this override, there would be only switching between the two first options in the list below. This adds the last option, if I understand correctly.
- Default in "con" - whatever the platform setting's logger type is, else WARNING
- Default in "txt" - whatever the platform setting's logger type is, else DEBUG
- With --verbose flag, disregard platform settings, always DEBUG in "con".

If I understand correctly, the "txt" file would remain unaffected by the use of --verbose flag. Why wouldn't we want the verbose override to apply to the file as well?

Javagedes:
> Overriding defaults when using --verbose makes sense to me. I think all debug messages should be accessible without getting into platform settings. Without this override, there would be only switching between the two first options in the list below. This adds the last option, if I understand correctly.
> 
> * Default in "con" - whatever the platform setting's logger type is, else WARNING
> * Default in "txt" - whatever the platform setting's logger type is, else DEBUG
> * With --verbose flag, disregard platform settings, always DEBUG in "con".
> 
> If I understand correctly, the "txt" file would remain unaffected by the use of --verbose flag. Why wouldn't we want the verbose override to apply to the file as well?

Hello! Your evaluation is correct. To answer your question in regard to why I didn't have the verbose flag override the file logger as well, is simply that I wanted to make the least number of changes. However, I think having the --verbose flag effect all loggers make sense, and is what I think most people would intuitively think should happen if opting in to verbose logging via the --verbose flag.

I will make the change! Thank you for the suggestion.

VivianNK (approved):

makubacki (approved):

PR #584, opened 2023-06-06
Update txt logger mode to 'a'

antklein:
Python logging to a file with mode='w' can cause log loss when logging happens after the logger has already been cleaned up. 
See issue: https://github.com/python/cpython/issues/86544

The fix for Python was to have a special handle for files opened in mode='w'. Currently we use mode='w+', so the checked in fix does not solve our problem.

I've implemented moving the file_logger to mode='a' and an explicit deletion of the logging file before opening the FileHandler.

### DIFF
--- a/edk2toolext/edk2_logging.py
+++ b/edk2toolext/edk2_logging.py
@@ -110,9 +110,14 @@ def setup_txt_logger(directory, filename="log", logging_level=logging.INFO,
     if not os.path.isdir(directory):
         os.makedirs(directory)
 
-    # Create file logger
     logfile_path = os.path.join(directory, filename + ".txt")
-    filelogger = file_handler.FileHandler(filename=(logfile_path), mode='w+')
+
+    # delete file before starting a new log
+    if os.path.isfile(logfile_path):
+        os.remove(logfile_path)
+
+    # Create file logger
+    filelogger = file_handler.FileHandler(filename=(logfile_path), mode='a')
     filelogger.setLevel(logging_level)
     filelogger.setFormatter(log_formatter)
     logger.addHandler(filelogger)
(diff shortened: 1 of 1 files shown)


antklein:
@makubacki looks like the Ubuntu Python 3.10 and 3.11 builds are complaining because of the below error. Who could help look into that?
```
Test Stuart PR for Policy 5 library inf changed                       | FAIL |
'' does not contain 'MdeModulePkg'
```

Javagedes:
> @makubacki looks like the Ubuntu Python 3.10 and 3.11 builds are complaining because of the below error. Who could help look into that?
> 
> ```
> Test Stuart PR for Policy 5 library inf changed                       | FAIL |
> '' does not contain 'MdeModulePkg'
> ```

Do not worry about this error. I added a new policy check that still need to be integrated into EDK2. This will continue to fail while I work through the mailing list and EDK2 PR process.

Javagedes:
@antklein This change looks good, but could you please specify the situation where this is an issue? The only way I see this being an issue is if you setup the same txt logger twice, which does not happen in an invocable.

Thanks!
Joey

antklein:
> 

Hi Joey, 
For us this exact issue was happening. https://github.com/python/cpython/issues/86544#issuecomment-1093891771

That's because in our platform level PlatformPostBuild function we spin off sub-threads which make use of the RunPythonScript from edk2toollib.utility_functions. If the main thread closes, closing the logging FileHandle, the sub-threads then return causing the RunPtyhonScript to log additional information clearing the original BUILDLOG output.

Javagedes:
> > @antklein This change looks good, but could you please specify the situation where this is an issue? The only way I see this being an issue is if you setup the same txt logger twice, which does not happen in an invocable.
> > Thanks! Joey
> 
> Hi Joey, For us this exact issue was happening. [python/cpython#86544 (comment)](https://github.com/python/cpython/issues/86544#issuecomment-1093891771)
> 
> That's because in our platform level PlatformPostBuild function we spin off sub-threads which make use of the RunPythonScript from edk2toollib.utility_functions. If the main thread closes, closing the logging FileHandle, the sub-threads then return causing the RunPtyhonScript to log additional information clearing the original BUILDLOG output.

Understood. That scenario makes sense. Thanks!

Javagedes (approved):

PR #605, opened 2023-07-12
nuget_dependency: authorization failure logging

Javagedes:
When a nuget package fails to download due to an authorization error, additional logging is provided to help the user understand the problem and possible solutions. The message informs the user to either install a credential provider (because the package requires authorization) or to run the command themselves to provide the credentials manually.

The logging appears as follows, per failed NuGet Package:
image

### DIFF
--- a/edk2toolext/environment/extdeptypes/nuget_dependency.py
+++ b/edk2toolext/environment/extdeptypes/nuget_dependency.py
@@ -248,7 +248,8 @@ class NugetDependency(ExternalDependency):
             if non_interactive and found_cred_provider:  # we should be interactive next time
                 self._attempt_nuget_install(install_dir, False)
             else:
-                if is_unauthorized:
+                # Only provide this error message if they are not using a credential provider, but receive a 401 error
+                if is_unauthorized and not found_cred_provider:
                     logging.warning("[Nuget] A package requires credentials, but you do not have a credential "\
                                     "provider installed.")
                     logging.warning("[Nuget] Please install a credential provider and try again or run the following "\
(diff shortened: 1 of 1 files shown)


edk2toolext/environment/extdeptypes/nuget_dependency.py
```diff
@@ -245,6 +248,12 @@ def _attempt_nuget_install(self, install_dir, non_interactive=True):
             if non_interactive and found_cred_provider:  # we should be interactive next time
                 self._attempt_nuget_install(install_dir, False)
             else:
+                if is_unauthorized:
+                    logging.warning("[Nuget] A package requires credentials, but you do not have a credential "\
```
makubacki, line 252: Is it possible they do have a credential provider installed but the credentials are incorrect?

Trying to make sure the message is not helpful but potentially misleading.

Javagedes: I've fixed the logic for when to print these warnings. These warnings are only for the situation where the user is not using a credential provider but is receiving a 401 error (unauthorized).

The scenario you mention above does not result in my message being printed. However out of curiosity, I did test what would happen if I logged into the wrong account for the credential provider. It just failed with the normal warning (`WARNING - [SDE] Failed to fetch NugetDependecy: ... `). On the second pass it pops up an error message allowing you to sign out, and sign back in with a different account.

makubacki: Thanks for checking, looks good with the latest commit.

makubacki (approved):

PR #654, opened 2023-09-19
Report parser sqlite3

Javagedes:
Adds new invocables for generating a database containing workspace information and reports to run against the database.

1. `stuart_parse` is a new invocable tied to a settings file that is used to parse workspace and environment information for a given package (or multiple packages). It creates a standalone sqlite3 database that can be queried using sql queries or reports managed by `stuart_report` can be run against the database.
2. `stuart_report` is a standalone executable that manages multiple reports that can be run against the database.

Stuart Parse can be run on a platform package (typically associated with a PlatformBuild.py file) or non-platform packages (typically associated with a CISettings.py file).

## Initial Reports

1. Coverage report: `stuart_report coverage` is a report that consumes a cobertura report generated from host based unit tests in the workspace, and re-organizes the report per INF rather than per executable to more accurately show test coverage.
2. Usage report: `stuart_report usage` is a report that generates an html document showing all INFs (both library and component) that are used to build a platform, and which submodule (or BASE repo) the INF comes from.
3. Component Library Report: `stuart_report component-libs` is a report that prints to the terminal (or a file) the actual list of library instances used to build a specific component. It can print in flat mode, or show a recursive list of dependencies for the component.


edk2toolext/environment/reporttypes/coverage_report.py
```diff
@@ -0,0 +1,337 @@
...
+                # Check if the file should be excluded
+                exclude_file = False
+                for pattern in self.args.exclude:
+                    if fnmatch.fnmatch(source, pattern):
+                        exclude_file = True
+                        break
+                if exclude_file:
+                    continue
+
+                if source in self.args.exclude:
```
cfernald, line 321: Is this not covered by the above pattern match search? If its a verbatim path then it should match as a pattern, no?

edk2toolext/environment/reporttypes/coverage_report.py
```diff
@@ -0,0 +1,337 @@
...
+            classes = ET.SubElement(inf, "classes")
+
+            for source in source_list:
+                # Check if the file should be excluded
+                exclude_file = False
+                for pattern in self.args.exclude:
+                    if fnmatch.fnmatch(source, pattern):
+                        exclude_file = True
+                        break
+                if exclude_file:
```
cfernald, line 318: should log the exclusion like you do below?

edk2toolext/environment/reporttypes/coverage_report.py
```diff
@@ -0,0 +1,339 @@
...
+            (bool): True if the report was successful, False otherwise.
+        """
+        # Verify valid ACTIVE_PLATFORM
+        dsc = self.args.dsc
+        if not dsc:
+            logging.error("No ACTIVE_PLATFORM dsc file specified. It should be edk2 package path relative.")
+            return -1
+        logging.info(f"ACTIVE_PLATFORM requested: {dsc}")
+
+        # replace any files with their contents
```
cfernald, line 169: This is interpreting the exclusion list as pointing to a file, but for the package, we dont do this. Is this intentionally different?

cfernald (approved):

edk2toolext/environment/reporttypes/coverage_report.py
Javagedes: Good catch. It was accidently left out of run_by_package. I'll move it up to run_report, so that both format types (package / platform) can consume the exclude list.

I'll also note that it supports a mixture of directly defining the pattern in the cli, providing a file of patterns, and a mixture.

i.e. `-e pattern1 -e file_with_patterns.txt -e pattern2` or `-e pattern1,file_with_patterns.txt,pattern2`

edk2toolext/environment/reporttypes/coverage_report.py
Javagedes: I agree. Removed. Thanks!

edk2toolext/environment/reporttypes/coverage_report.py
Javagedes: Added logging. Thanks!

edk2toolext/environment/reporttypes/coverage_report.py
Javagedes: Created a function and moved its call to `run_report`. Thanks!

Issue #658, opened 2023-09-22
[Bug]: Better recognize cargo/cargo-make errors


makubacki:
### Describe the Bug

We recently encountered an issue where `cargo make` errors did not get noticed for a while because the build continued.

For example:

```
  [cargo-make] Execute Command: "cargo" "tarpaulin" "-p" \
    "HelloWorldRustDxe,RustBootServicesAllocatorDxe"

  cargo_tarpaulin::config: Creating config
  cargo_tarpaulin: Running Tarpaulin
  cargo_tarpaulin: Building project
  cargo_tarpaulin::cargo: Cleaning project
  error: invalid character `,` in pkgid:
    `HelloWorldRustDxe,RustBootServicesAllocatorDxe`, characters must
    be Unicode XID characters (numbers, `-`, `_`, or most letters)
  cargo_tarpaulin: Cargo failed to run! Error: cargo run failed
  Error: "Cargo failed to run! Error: cargo run failed"
  [cargo-make] ERROR - Error while executing command, exit code: 1
  [cargo-make] WARN - Build Failed.
```

### What Python version are you using?

Python 3.11

### Reproduction steps

Review the `MsCorePkg` pipeline output as of commit `dec8f57d3a5dc3f270f6b18940e99ba114d25f7a`.

### Expected behavior

It would be helpful for the build to clearly identify the errors.

### Execution Environment

- Mu Plus at commit
- Rust 1.71.1
- Cargo-make 0.37.1
- Cargo-tarpaulin 0.27.0
- Build agent details
  - https://github.com/actions/runner-images/blob/win22/20230918.1/images/win/Windows2022-Readme.md
  - https://github.com/actions/runner-images/releases/tag/win22%2F20230918.1

Javagedes:
This is to improve detecting and logging rust build errors as errors so that the developer does not need to parse through the build log. The changes would be in https://github.com/tianocore/edk2-pytool-extensions/blob/5acb5d1cc96d2bf774579607c1c5cc960d3563b9/edk2toolext/edk2_logging.py#L208C5-L208C25

PR #665, opened 2023-09-26
edk2toolext/edk2_logging.py: Log rust errors

Javagedes:
Adds the ability to detect rust build errors and print them at the end of the build when the build fails.

image

### DIFF
--- a/edk2toolext/edk2_logging.py
+++ b/edk2toolext/edk2_logging.py
@@ -231,6 +231,7 @@ def scan_compiler_output(output_stream):
     build_py_error_exp = re.compile(r"error (\d+)E:")
     linker_error_exp = re.compile(r"error LNK(\d+):")
     warning_exp = re.compile(r"warning [A-Z]?(\d+):")
+    rust_error_exp = re.compile(r"^(-->|\d+\s*\||error:)")
     for raw_line in output_stream.readlines():
         line = raw_line.strip("\n").strip()
         match = error_exp.search(line)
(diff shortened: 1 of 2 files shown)


Javagedes:
@makubacki is this what you were requesting in #658? We don't currently try to catch the c-style "clippy-like" explanations when compiles fail. See test `test_scan_compiler_output_gcc_mixed_actual`.

I know on the mu_basecore side of things you want to catch silent failures in the plugin, but not 100% sure what you are looking for on the extensions side of things as described in #658

makubacki:
> @makubacki is this what you were requesting in #658? We don't currently try to catch the c-style "clippy-like" explanations when compiles fail. See test `test_scan_compiler_output_gcc_mixed_actual`.
> 
> I know on the mu_basecore side of things you want to catch silent failures in the plugin, but not 100% sure what you are looking for on the extensions side of things as described in #658

I was looking to add more context in error messages so problems can better be understood. The output shown in the PR description looks fine to me.

edk2toolext/edk2_logging.py
```diff
@@ -231,6 +231,7 @@ def output_compiler_error(match, line, start_txt="Compiler"):
     build_py_error_exp = re.compile(r"error (\d+)E:")
     linker_error_exp = re.compile(r"error LNK(\d+):")
     warning_exp = re.compile(r"warning [A-Z]?(\d+):")
+    rust_error_exp = re.compile(r"^(-->|\d+\s*\||error:)")
```
makubacki, line 234: Could you add a few tests?

Javagedes: the regex was reduced to only `error:`, `error[errotype]:` and `-->` to reduce conflicts with catching unexpected c code. Example in the PR summary. 

Tests have been added. Thanks!

makubacki (approved):

edk2toolext/edk2_logging.py
Javagedes: I have added tests. Thanks!

PR #678, opened 2023-10-20
Add Rust tool version reporting

makubacki:
Adds support to report Rust tool versions in the version aggregator so the versions are available in the Build Tools Report.

Also prints the version to the logger at info level similar to other versions so it is part of the build output.

---

Example output in the build console:

```
INFO - cargo version: 1.73.0
INFO - cargo make version: 0.37.2
INFO - rustc version: 1.73.0
```

Example output in the HTML build report:
image

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -114,6 +114,7 @@
         "rmtree",
         "rpartition",
         "rtags",
+        "rustc",
         "schtasks",
         "sdist",
         "setuplog",
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -28,7 +28,7 @@ from textwrap import dedent
 from typing import Iterable, Tuple
 
 import pkg_resources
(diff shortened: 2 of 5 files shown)


Javagedes:
Could you write a test that makes sure we parse cargo and rustc versions as expected? Just to ensure if the executables start reporting their versions differently, we catch the break here. We don't have a great way to do it with cargo make unless we go through installing it first, which is slow. It may or may not be worth it.

makubacki:
> Could you write a test that makes sure we parse cargo and rustc versions as expected? Just to ensure if the executables start reporting their versions differently, we catch the break here. We don't have a great way to do it with cargo make unless we go through installing it first, which is slow. It may or may not be worth it.

Sure, I'll do that tomorrow. Thanks for quickly looking at the PR.

makubacki:
> > Could you write a test that makes sure we parse cargo and rustc versions as expected?

> Sure, I'll do that tomorrow. Thanks for quickly looking at the PR.

Done in the latest push.

Javagedes (approved):

PR #690, opened 2023-11-10
coverage_report: Add additional functionality

Javagedes:
Adds the additional functionality to the coverage report:

1. Generate xml data for source files that do not have coverage data with the `--full` flag. Using this flag requires the `-ws` flag is pointing to the workspace root, so that the files can be found for code counting.
2. Adds the ability to "flatten"reports with the `--flatten` flag. This flag causes the report to place all source files together, rather than grouping them by INF. This results in different coverage data as duplicate sources files are removed (which can occur when a certain source file is used by multiple INFs). 
3. Changes the `name` attribute for a file to be the package path relative path. This is because report generator tools that consume the coverage.xml file combine coverage data for source files with the same `name` attribute, even if the `filepath` attributes are different.
4. writes the coverage xml in "pretty" format, which causes the xml to have the expected indentation instead of no newlines whatsoever. This has no effect, but makes reading the coverage.xml file easier.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -104,6 +104,7 @@
         "prebuild",
         "prepended",
         "pygit",
+        "pygout",
         "pymodule",
         "pyopenssl",
         "pypath",
--- a/.cspell.json
+++ b/.cspell.json
@@ -104,7 +104,7 @@
         "prebuild",
         "prepended",
         "pygit",
-        "pygout",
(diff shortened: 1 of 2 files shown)


cfernald (approved):

Javagedes:
@dependabot rebase

PR #826, opened 2024-06-06
Add Linux/Windows variable support

apop5:
Adding Uefi Variable Access to pytool-extensions.

Multiple copies of this support are floating in repos, moving to pytool-extensions as the first step.
Follow on steps will modify existing scripts to use consolidated version.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -148,5 +148,10 @@
         "vsts",
         "xffd",
         "xlsxwriter",
+        "uefivariablesupport",
+        "ntdll",
+        "NTSTATUS",
+        "efivarfs",
+        "chattr"
     ]
 }
--- a/edk2toolext/os/__init__.py
+++ b/edk2toolext/os/__init__.py
@@ -1,0 +1,6 @@
+##
+# Copyright (c) Microsoft Corporation
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+"""A python class to allow interaction with Uefi Variables from an OS."""
--- a/edk2toolext/os/uefivariablesupport.py
+++ b/edk2toolext/os/uefivariablesupport.py
@@ -1,0 +1,331 @@
+# @file uefivariablesupport.py
+#
+# Exports Class to allow OS level interaction
+# with UEFI variables. Includes Windows and
+# Linux support.
+#
+#
+# Copyright (c), Microsoft Corporation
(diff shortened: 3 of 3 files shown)


edk2toolext/os/UefiVariableSupportLib.py
```diff
@@ -0,0 +1,310 @@
...
+import os
+import uuid
+import struct
+import logging
+
+from ctypes import (
+    create_string_buffer
+)
+
+if os.name == 'nt':
```
Javagedes, line 24: I see that you've already separated out the windows logic and os agnostic logic, which is awesome. I think, however, that since this only works on windows currently, just to add an else statement to this initial check (not the others) that will raise a `NotImplementedError` stating `os.name` is not currently supported for this module.

Thoughts?

edk2toolext/os/UefiVariableSupportLib.py
```diff
@@ -0,0 +1,310 @@
...
+    # Function to get all variable names
+    # return a tuple of error code and variable names byte array formatted as:
+    #
+    # typedef struct _VARIABLE_NAME {
+    #   ULONG NextEntryOffset;
+    #   GUID VendorGuid;
+    #   WCHAR Name[ANYSIZE_ARRAY];
+    # } VARIABLE_NAME, *PVARIABLE_NAME;
+    #
+    def GetUefiAllVarNames(self):
```
Javagedes, line 161: As ruff has yelled at you already, we use a different format for function comments. You can see other modules :)

edk2toolext/os/UefiVariableSupportLib.py
apop5: This should contain the logic to support both windows and Linux.

Its joining both the windows and linux implementations under a single class to try to simplify for the consumers to be able to add a single class for the functionality.

If you have alternative ideas of how to do this, I would appreciate any pointers.

Javagedes: I missed that it was support for both ~ It's fine the way it is then :)

edk2toolext/os/UefiVariableSupportLib.py
apop5: should be addressed in latest.

apop5:
Based on advice, will abandon this PR and move functionality into [edk2-pytool-library] (https://github.com/tianocore/edk2-pytool-library/pull/577)

PR #871, opened 2024-08-22
nuget_dependency: Allow path override

Javagedes:
As described in [external_dependency.py#L38](https://github.com/tianocore/edk2-pytool-extensions/blob/1e4ddee16145c262b96c67bd442124563400d8d6/edk2toolext/environment/external_dependency.py#L38), the `name` configuration is the name of the ext_dep, used to name the folder the ext_dep will be unpacked into. When the nuget external dependency was added, `name` also became the name of the package you are downloading, which tightly couples the name of the package and the folder path. This means there is no way to name the folder something other than the package name, which can be necessary in some situations.

This change provides a backwards compatible way to de-couple the package name from the folder name by introducing a new, optional config for nuget dependencies, `package`. `package` will be used as the package name to download while `name` will continue to be the name of the ext dep and the folder it is downloaded to. For backwards compatability, if `package` is not used in the configuration file, `name` will be used instead.

### DIFF
--- a/.github/workflows/VariableProducer.yml
+++ b/.github/workflows/VariableProducer.yml
@@ -14,7 +14,7 @@ on:
         value: ${{ jobs.produce.outputs.node-versions }}
 
 env:
-  pythonversions: "['3.12', '3.11', '3.10']" # Keep Python Versions in descending order
+  pythonversions: "['3.12.4', '3.11', '3.10']" # Keep Python Versions in descending order
   nodeversions: "['19']"
 
 jobs:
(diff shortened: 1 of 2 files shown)


apop5:
Is there an example for this?

apop5 (approved):

Javagedes:
> Is there an example for this?

@apop5  The example that brought about this requested change is header files that are downloaded via external dependency and are in a package's `Include/*` folder. There are two mutually exclusive external dependencies at this location, but only one exists at any point in time. In that package's dec file, you cannot simply add both folders to it, because one will exist and one will not, and the build will fail. So they need to be able to download one or the other extdep to the same folder.

Additionally, macros don't work in dec files, so we cannot set a build variable within the extedep then do something along the lines of:

```toml
[Includes]
$(VAR_NAME)
```

PR #966, opened 2025-03-31
Improve configuration options for GetDependencies.

apop5:
GetRequiredSubmodules() vs GetDependencies()

GetRequiredSubmodules() allows a recursive/omnicache configuration parameter to be passed. 
The configuration parameter determines if --recursive is passed when the submodule is cloned.

GetDependencies() does not have configuration for this type of scenario.
A Dependency from GetDependencies will always attempt to resolve **ALL** submodules that are contained in the repo.

This presents a problem when a submodule contains their own dependency as well. The example run into is:

SecurityPkg => libspdm
libspdm => openssl
openssl => gost-engine
gost-engine => libprov

```
{
	"Path": "Common/MU_TIANO",
	"Url": "https://github.com/microsoft/mu_tiano_plus.git",
	"Branch": "dev/202502",
},
```

```
{
	"Path": "Common/MU_TIANO",
	"Url": "https://github.com/microsoft/mu_tiano_plus.git",
	"Branch": "dev/202502",
	"Recurse": True
},
```

```
{
	"Path": "Common/MU_TIANO",
	"Url": "https://github.com/microsoft/mu_tiano_plus.git",
	"Branch": "dev/202502",
	"Recurse": False
},
```
```
{
	"Path": "Common/MU_TIANO",
	"Url": "https://github.com/microsoft/mu_tiano_plus.git",
	"Branch": "dev/202502",
	"Recurse": {"CIFile" : ".pytool/CISettings.py"}
},
```

Add "Recurse" support for GetDependencies function.
Possible values are True/False/Dictionary that Specifies a CIFile.
If a CI file is specified, the GetRequiredSubmodules of the CI file will be
used to determine which submodules should be cloned.

### DIFF
--- a/edk2toolext/environment/repo_resolver.py
+++ b/edk2toolext/environment/repo_resolver.py
@@ -372,13 +372,10 @@ def clone_repo(abs_file_system_path: os.PathLike, DepObj: dict) -> tuple:
         branch = DepObj["Branch"]
     if "ReferencePath" in DepObj and os.path.exists(DepObj["ReferencePath"]):
         reference = Path(DepObj["ReferencePath"])
-    if "Recurse" in DepObj:
-        if type(DepObj["Recurse"]) is not dict:
-            recurse = DepObj["Recurse"]
-        else:
-            recurse = False
+    if type(DepObj["Recurse"]) is not dict:
+        recurse = DepObj["Recurse"]
     else:
-        recurse = True
+        recurse = False
 
     # Used to generate clone params from flags
     def _build_params_list(branch: str = None, shallow: str = None, reference: str = None, recurse: str = None) -> None:
@@ -451,17 +448,14 @@ def checkout(
         ci_file = None
         if "ReferencePath" in dep and os.path.exists(dep["ReferencePath"]):
             reference = Path(dep["ReferencePath"])
-        if "Recurse" in dep:
-            if type(dep["Recurse"]) is not dict:
-                recurse = dep["Recurse"]
-            else:
-                recurse = False
-                try:
-                    ci_file = dep["Recurse"]["CIFile"]
-                except KeyError:
-                    logging.error(f"Failed to find 'CIFile' 'Recurse' section {dep['Recurse']}")
+        if type(dep["Recurse"]) is not dict:
+            recurse = dep["Recurse"]
         else:
-            recurse = True
+            recurse = False
+            try:
+                ci_file = dep["Recurse"]["CIFile"]
+            except KeyError:
+                logging.error(f"Failed to find 'CIFile' 'Recurse' section {dep['Recurse']}")
 
         if "Commit" in dep:
             commit = dep["Commit"]
--- a/edk2toolext/environment/repo_resolver.py
+++ b/edk2toolext/environment/repo_resolver.py
@@ -15,6 +15,7 @@ The intent is to keep all git functionality consolidated in this module. Current
 edk2_setup.py, and git_dependency.py use this module to perform git operations.
 """
 
+import importlib
 import logging
 import os
 import time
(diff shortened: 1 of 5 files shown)


edk2toolext/environment/repo_resolver.py
```diff
@@ -420,6 +438,34 @@ def checkout(
                 else:
                     logger.critical(f"Dependency {dep['Path']} is not in sync with requested commit.  Fail.")
                     raise Exception(f"Dependency {dep['Path']} is not in sync with requested commit.  Fail.")
+
+            if recurse is None:
+                for a in Path(abs_file_system_path).glob("**/CISettings.py"):
```
Javagedes, line 443: We do not force that the CI file is named "CISettings.py". It can be called anything and located anywhere. I also think there could be a scenario where we do not want to follow what the lower level CISettings.py file does.

I think the better way is to have a nested config in the CISettings.py file that is being executed. Something like

## Normal way, untouched

``` json
{
    "Path": "Common/MU_PLUS",
    "Url": "https://github.com/Microsoft/mu_tiano_plus",
    "Branch": "release/202405"
}
```

## New way, but same logic as normal way

``` json
{
    "Path": "Common/MU_PLUS",
    "Url": "https://github.com/Microsoft/mu_tiano_plus",
    "Branch": "release/202405"
    "Recurse": true,
}
```

## Stop all recursion

``` json
{
    "Path": "Common/MU_PLUS",
    "Url": "https://github.com/Microsoft/mu_tiano_plus",
    "Branch": "release/202405"
    "Recurse": false,
}
```

## Only recurse the specified submodules

In this example, lets imagine mu_tiano_plus has a bunch of submodules, but we only want to clone libspdm (opposite of what you actually want), then we would do this:

``` json
{
    "Path": "Common/MU_PLUS",
    "Url": "https://github.com/Microsoft/mu_tiano_plus",
    "Branch": "release/202405"
    "Recurse": [
        {
            "Path": "SecurityPkg/DeviceSecurity/SpdmLib/libspdm",
            "Url": "https://github.com/DMTF/libspdm.git",
            "Commit": "abcdef"
        }
    ]
}
```

apop5: I am not opposed, but we are going to bleed configuration from one repo to the next, and it becomes a Maintenace problem.

Two real world examples

libspdm version was updated. Once the update goes into MU_PLUS, then every repo and project that depends on  it will also require being updated with the information.

ArmSoftFloatLib, it was removed. All of the repos that were consuming mu_basecore would be required to go update their CISettings file to remove the ArmSoftFloatLib

apop5: After talking with Sean, settled on the course of action of having the GetDependencies provide a path to the CI file, if it exists. If CIFile is empty, no submodules will be cloned, if it exists, only submodules in file will be cloned, and if not specified at all, recurse all submodules. 

{
    "Path": "Common/MU_PLUS",
    "Url": "https://github.com/Microsoft/mu_tiano_plus",
    "Branch": "release/202405"
    "CIFile": ".pytool/CISettings.py"
}

Javagedes: Good idea :)

edk2toolext/environment/repo_resolver.py
```diff
@@ -420,6 +438,35 @@ def checkout(
                 else:
                     logger.critical(f"Dependency {dep['Path']} is not in sync with requested commit.  Fail.")
                     raise Exception(f"Dependency {dep['Path']} is not in sync with requested commit.  Fail.")
+
+            if not recurse:
+                CISettingFile = Path(abs_file_system_path, dep["CIFile"])
```
Javagedes, line 443: Please throw this in a support function in this file. :)

Javagedes (approved):

PR #969, opened 2025-04-08
edk2toolext: Add patch var store [Rebase & FF]

makubacki:
Adds a script that allows the UEFI variable store in a ROM file to be patched with variable data specified in an XML file.

---

Another commit moves all integration testing to Ubuntu 24.04 since the 22.04 container image is being deprecated.


Javagedes:
@makubacki What is the use-case for this being a standalone script? It seems like it will be more beneficial as a helper plugin.

Also, if you want it to be registered and callable from the command line, you'll want to register it here:

https://github.com/tianocore/edk2-pytool-extensions/blob/026dab581aee8c2d27aec82d6119699bcbc81664/pyproject.toml#L73

makubacki:
> Also, if you want it to be registered and callable from the command line, you'll want to register it

Done

.cspell.json
```diff
@@ -78,6 +78,7 @@
         "keyserver",
         "levelno",
         "linuxize",
+        "manulipulations",
```
joschock, line 81: heh. seems like we should file an issue to get this fixed in the edk2 tool-lib. Not a blocker for this PR, given that it's caused by upstream naming.

joschock (approved):

.cspell.json
makubacki: Yeah, it was disappointing to see that. Tracking issue filed here - https://github.com/tianocore/edk2-pytool-library/issues/724. Resolving this thread.

Issue #971, opened 2025-04-21
[Bug]: repo_resolver.py - _build_params_list removed the else statement for appending "--recurse-submodules" to the git clone command


antklein:
### Contact Details

antklein@microsoft.com

### Describe the Bug

stuart_ci_setup fails to clone recursive submodules when the `git clone` reference is an edk2-pytool-extensions omnicache reference.

The `--recurse-submodules` feature of git clone expects the reference passed to be in a specific format for the operation to function properly. The omnicache feature does not meet this format requirement, causing the submodule clone to fail.

Reference to git submodule requirements: https://github.com/git/git/blob/4bbb303af69990ccd05fe3a2eb58a1ce036f8220/builtin/submodule--helper.c#L1590

### What Python version are you using?

Python 3.13

### Reproduction steps

1. Initialize edk2-pytool-extensions omnicache for the repository.
2. Run stuart_ci_setup passing in the omnicache path.
3. Setup will fail to recursively clone submodules for CI dependencies.

### Expected behavior

1. Initialize edk2-pytool-extensions omnicache for the repository.
2. Run stuart_ci_setup passing in the omnicache path.
3. Setup succeeds to recursively clone all submodules for CI dependencies.

### Additional context

Failure log snippet:

```
INFO - Cloning at C:\w\1\s\MU_BASECORE
INFO - Cloning repo: https://github.com/microsoft/mu_basecore.git
DEBUG - Popen(['git', 'clone', '-v', '--reference', 'C:/w/OmniCache', '--recurse-submodules', '--', 'https://github.com/microsoft/mu_basecore.git', 'C:\\w\\1\\s\\MU_BASECORE'], cwd=C:\w\1\s, stdin=None, shell=False, universal_newlines=True)
```

```
hint: An alternate computed from a superproject's alternate is invalid.\n
hint: To allow Git to clone without an alternate in such a case, set\n
hint: submodule.alternateErrorStrategy to 'info' or, equivalently, clone with\n
hint: '--reference-if-able' instead of '--reference'.\n

fatal: submodule 'BaseTools/Source/C/BrotliCompress/brotli' cannot add alternate: path 'C:/w/OmniCache/modules/BaseTools/Source/C/BrotliCompress/brotli/' does not exist\n
Failed to clone 'BaseTools/Source/C/BrotliCompress/brotli'. Retry scheduled\n
```

PR #972, opened 2025-04-21
repo_resolver: fix case of reference and recurse being mutually exclusive.

apop5:
--reference and --recurse-submodules cannot be used at the same time.

Resolve issue introduced in #966 that allowed the parameter to be used at the same time during git clone with omnicache.


apop5:
Fixes #971

Javagedes (approved):

PR #979, opened 2025-05-16
Adding limited cloning capabilities to stuart_setup

apop5:
Use case:

In a PlatformBuild.py file, GetRequiredSubmodules() is implemented to list
the required submodules that need to be downloaded for the platform build.

These submodules, today, will only have the True/False configuration option, which will 
recurse all submodules, including the submodules submodules.
(Example: mu_tiano_plus includes LibSpdm, which includes submodule openssl, which includes submodule pyca/cryptography)
But, repo dependent, the submodules of a submodule are not required to build/use the functionality.

Add an optional configuration option that will allow specifying a CISettings file and only the submodules in the settings file will be downloaded.

Test Environment:
Modify mu_tiano_platform's PlatformBuild.py:

`RequiredSubmodule("MU_BASECORE", True,),
RequiredSubmodule("Common/MU", True,),
RequiredSubmodule("Common/MU_TIANO", True, ".pytool/CISettings.py"),
RequiredSubmodule("Common/MU_OEM_SAMPLE", True,),
RequiredSubmodule("Features/DEBUGGER", True,),
RequiredSubmodule("Features/DFCI", True,),
RequiredSubmodule("Features/CONFIG", True,),
RequiredSubmodule("Features/MM_SUPV", False),`

run stuart_setup and then verify that submodules of mu_tiano_plus were not recursively downloaded.

### DIFF
--- a/edk2toolext/environment/repo_resolver.py
+++ b/edk2toolext/environment/repo_resolver.py
@@ -316,6 +316,7 @@ def process_submodules_from_ci_file(
             CiSettingsInstance = importlib.util.spec_from_file_location(module_name, CISettingFile)
             Settings = importlib.util.module_from_spec(CiSettingsInstance)
             CiSettingsInstance.loader.exec_module(Settings)
+
             workspace_path = Settings.Settings().GetWorkspaceRoot()
             submodule_list = Settings.Settings().GetRequiredSubmodules()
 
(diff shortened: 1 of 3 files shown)


edk2toolext/environment/repo_resolver.py
```diff
@@ -598,15 +600,18 @@ def submodule_resolve(
         repo.git.submodule("sync", "--", submodule.path)
 
         params = ["update", "--init"]
-        if submodule.recursive:
+        if submodule.recursive and not submodule.configuration_file:
```
Javagedes, line 603: I would argue that `submodule.recursive` and `submodule.configuration_file` both being set by the consumer should be a error and exit or at least a warning, rather than quietly ignoring the recursive flag.

Any reason I'm thinking wrong with that?

apop5: I was trying to user proof it.

If the user was making a change and missed changing recursive to false but provided a config file, I was defering to config file.

There are a lot more cases of users using True today then using the config file....

Javagedes: haha fair enough. :)

Javagedes (approved):

PR #980, opened 2025-05-16
environment/repo_resolved: Support branches from gh pr checkout

mdkinney:
When checking out a PR using the gh command, the tracking branch may not be associated with any of the current remotes. This causes a python exception determining the URL of the remote associated with the PR branch when running stuart commands.

However, the URL to the repo the PR was created from can be determined by parsing the tracking_branch() and removing the branch name from the end of the tracking_branch().

### DIFF
--- a/edk2toolext/environment/repo_resolver.py
+++ b/edk2toolext/environment/repo_resolver.py
@@ -274,9 +274,14 @@ def repo_details(abs_file_system_path: os.PathLike) -> dict:
             details["Remotes"] = [remote.name for remote in repo.remotes]
 
             # 1. Use the remote associated with the branch if on a branch and it exists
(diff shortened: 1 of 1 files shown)


apop5:
@mdkinney 

Can you give the example command that you used for testing this?

Javagedes:
> ~~@mdkinney ~~
> 
> ~Can you give the example command that you used for testing this?~
> 
> Joey explained it, you are checking out the PR and then running stuart_setup to hydrate the submodules. Your changes make sense.
> 
> The ruff error just needs fixed.
> 
> You can run `ruff format .` and it have it format the files in place. (python -m pip install ruff, if its not already installed)

@apop5 After our conversation I looked more, and we use the repo resolver in the self-describing environment, which runs in all invocables. So he is probably also just hitting this with stuart_ci_build

Javagedes (approved):

PR #986, opened 2025-05-30
nuget_dependency: allow source override

gonzoua:
Introduce an env variable NUGET_PKG_SOURCE that if set would override source for NuGet packages. This allows to install NuGet dependencies in restricted environments without global network access by pre-caching them in local directory or network share.

### DIFF
--- a/docs/user/features/extdep.md
+++ b/docs/user/features/extdep.md
@@ -73,6 +73,10 @@ This package has a version of the NuGet.exe binary or a user can configure their
 environment to use a defined version by setting `NUGET_PATH` to the folder containing
 the NuGet.exe that should be used.
 
+For enviroments with restricted network access it's possible to override
+dependencies' sources with a single local mirror by setting the value of
+the `NUGET_PKG_SOURCE` env variable to the mirror's location.
+
 ### Web Dependency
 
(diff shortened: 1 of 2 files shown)


Javagedes:
Hello @gonzoua, similar functionality already exists, however it is more broadly usable for all external dependency types. This functionality allows you to set a local path (`STUART_EXTDEP_CACHE_PATH`), which external dependencies attempt to pull from first, before reaching out over the network to download. Can you investigate this for your use case before we introduce new functionality?

Here are some docs regarding it, and I can help answer any questions you may have.

[stuart_extdep_cache_path](https://www.tianocore.org/edk2-pytool-extensions/features/environment_variables/#stuart_extdep_cache_path)
[local-caching](https://www.tianocore.org/edk2-pytool-extensions/features/extdep/#local-caching)

And here is where it is used in code:

https://github.com/tianocore/edk2-pytool-extensions/blob/1d6c1767027394097675336e4ec1d757feccbbb9/edk2toolext/environment/external_dependency.py#L153-L154

gonzoua:
Hi @Javagedes, thank you for references. Both approaches work in my environment. I'll close this pull request.

PR #989, opened 2025-06-05
edk2_invocable: Register one version per pip package

Javagedes:
This commit results in only the first version of a pip package found on the path being used as the version, even if multiple versions exist in the distribution list because it exists in multiple path locations. If using a venv environment, the path to those pip modules is prepended, so it should theoretically always be correct.

### DIFF
--- a/edk2toolext/edk2_invocable.py
+++ b/edk2toolext/edk2_invocable.py
@@ -209,13 +209,14 @@ class Edk2Invocable(BaseAbstractInvocable):
             pip_packages = []
         # go through all installed pip versions
         for package in pip_packages:
-            try:
-                version = package.version
-                name = package.metadata["Name"] if "Name" in package.metadata else package.metadata.get("name", None)
-                if not name:
-                    name = package.metadata.get("Summary", "unknown")
-            except Exception:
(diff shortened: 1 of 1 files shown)


edk2toolext/edk2_invocable.py
```diff
@@ -209,13 +209,14 @@ def collect_python_pip_info(cls: "Edk2Invocable") -> None:
...
-            except Exception:
+            name = package.metadata.get("Name", None) or package.metadata.get("Summary", None)
+            if name is None:
                 continue
+
+            # Always grab the version of the first instance of the package on the PATH. pipenv inserts its
+            # path first, so we can assume that the first instance is the one actually used.
+            version = importlib.metadata.version(name)
+
             logging.info("{0} version: {1}".format(name, version))
```
antklein, line 220: Nitpick: This will still print duplicate pip modules in the build output. This could be filtered to prevent that.
Example - 
BUILDLOG.txt
278: INFO - typing_extensions version: 4.14.0
293: INFO - typing_extensions version: 4.14.0

This should not be considered a blocking issue.

antklein (approved):

apop5 (approved):

PR #998, opened 2025-07-01
Sort the plugins by the optional order attribute

v-yuanhengwu:
If order is not specified in the plugin descriptor, apply a default value 0. That makes this code diff backward compatible since python sorts should be stable.

"Sorts are guaranteed to be stable. That means that when multiple records have the same key, their original order is preserved."

Refer to:
https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts

### DIFF
--- a/edk2toolext/environment/plugin_manager.py
+++ b/edk2toolext/environment/plugin_manager.py
@@ -32,6 +32,7 @@ class PluginDescriptor(object):
         self.Obj = None
         self.Name = t["name"]
         self.Module = t["module"]
+        self.Order = t.get("order", 0)
 
     def __str__(self) -> str:
         """String representation of the plugin descriptor."""
@@ -65,6 +66,7 @@ class PluginManager(object):
                 self.Descriptors.append(b)
             else:
                 failed.append(a)
+        self._sortPlugins()
         return failed
 
     def GetPluginsOfClass(self, classobj: type) -> list[object]:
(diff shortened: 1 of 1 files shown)


Javagedes:
Hello,

Thank you for the contribution. However, we intentionally do not guarantee plugin execution order. This is a deliberate design choice to ensure plugins remain decoupled and independently usable. Plugins are designed to be drop-in components without dependencies on other plugins. This approach allows platforms to scope and register only the plugins they need without having to account for execution order or hidden dependencies.

Please see this documentation: [Types of Plugins](https://www.tianocore.org/edk2-pytool-extensions/features/plugin_manager/#types-of-plugins)

image

If there's a specific scenario you're trying to address, I'm happy to discuss it further in a GitHub issue or discussion thread and work together on a fair solution. However, at this time, we intend to continue with the design decision not to guarantee plugin execution order.

spbrogan:
Adding additional context: 
The "Self-Describing Environment" was designed for a world where only the platform layer is all knowing.  Plugins that are distributed at the other layers (core, silicon, features, etc) cannot and do not know what other layers are included, therefore it is impossible to create an ordering system that holds to this principle and doesn't start to make assumptions or requirements that may not be true for the platform.  

More reading is here: 

https://www.tianocore.org/edk2-pytool-extensions/#the-environment
https://www.tianocore.org/edk2-pytool-extensions/features/sde/

v-yuanhengwu:
I got it. Thanks for your explanation.
Yes we are working on a scenario and found that we need to execute PluginA prior to PluginB (both of them are do_pre_build). But I think we could try to find another approach first.
Shall we close this Pull Request?

PR #1000, opened 2025-07-01
edktoolext: Make BUILD_OUT_TEMP and BUILD_OUTPUT_BASE overridable

anandhkrishnau:
This commit modifies the default behavior of BUILD_OUT_TEMP and BUILD_OUTPUT_BASE variables to allow them to be easily overridden by users.

Previously, these variables had limited
override mechanisms, which made it difficult for users to customize their build output directories without modifying the core tool's source.

### DIFF
--- a/edk2toolext/environment/uefi_build.py
+++ b/edk2toolext/environment/uefi_build.py
@@ -569,6 +569,7 @@ class UefiBuilder(object):
             "BUILD_OUT_TEMP",
             os.path.normpath(os.path.join(self.ws, self.env.GetValue("OUTPUT_DIRECTORY"))),
             "Computed in SetEnv",
+            True
         )
 
         target = self.env.GetValue("TARGET", None)
@@ -581,6 +582,7 @@ class UefiBuilder(object):
             "BUILD_OUTPUT_BASE",
             os.path.join(self.env.GetValue("BUILD_OUT_TEMP"), target + "_" + self.env.GetValue("TOOL_CHAIN_TAG")),
             "Computed in SetEnv",
+            True
         )
 
         # We have our build target now.  Give platform build one more chance for target specific settings.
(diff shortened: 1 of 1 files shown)


anandhkrishnau:
@Javagedes Could you please help review this PR

anandhkrishnau:
@makubacki could you please help review this PR

Javagedes:
Hi @anandhkrishnau,

Apologies for the delayed response—I needed time to consult with the rest of the team.

After discussion, we’ve agreed that stuart_build is not designed to support running multiple EDK II build.py commands in a single invocation. The self-describing environment is intended to configure and execute a single platform build per run. Supporting multiple builds introduces several challenges, including (but not limited to):

1. Unexpected or inconsistent environment variable values between builds
1. Plugins that are expected to run once per build may not execute correctly
1. Dependencies in Stuart on specific environment variables that, when changed mid-run, can lead to corrupted pre- and post-build logic

To better illustrate the complexity, you can take a look at the implementation of stuart_ci_build, where we work around these limitations to support CI across different package/architecture/target combinations:
[stuart_ci_build](https://github.com/tianocore/edk2-pytool-extensions/blob/3584b5bc2e0fda414b12d459d1fa26627a968364/edk2toolext/invocables/edk2_ci_build.py#L183).

There, we explicitly snapshot and restore variable state to ensure isolation between builds.

Because of this, we do not officially support multiple builds within a single stuart_build command. In fact, certain environment variables are explicitly marked as non-overridable to help prevent accidental misuse that could result in undefined behavior.

That said, if your use case requires this behavior and you’re comfortable accepting the risks, we do provide a mechanism to override protected variables. You can use the [AllowOverride](https://github.com/tianocore/edk2-pytool-extensions/blob/3584b5bc2e0fda414b12d459d1fa26627a968364/edk2toolext/environment/var_dict.py#L170-L171) method to explicitly permit the override.

Your code would look like this:

```python
self.env.AllowOverride("BUILD_OUTPUT_BASE")
self.env.SetValue("BUILD_OUTPUT_BASE", ...)
```

Thanks again for your contribution and understanding.

PR #1003, opened 2025-07-02
stuart_ci_build: Add support for listing plugins

Javagedes:
This commit adds a new cli argument, `--list` to stuart_ci_build that will list all available CI plugins and the path to their source / documentation directory.

example of the output of `stuart_ci_build -c .pytool/CISetting.py`

image

### DIFF
--- a/edk2toolext/invocables/edk2_ci_build.py
+++ b/edk2toolext/invocables/edk2_ci_build.py
@@ -119,7 +119,7 @@ class Edk2CiBuild(Edk2MultiPkgAwareInvocable):
         """Retrieve command line options from the argparser."""
         self.disable_plugins = args.disable
         self.fail_fast = args.fail_fast
-        self.list_plugins = args.list
+        self.list_plugins = args.plugin_list
         super().RetrieveCommandLineOptions(args)
 
     def GetSettingsClass(self) -> type:


edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -105,12 +106,20 @@ def AddCommandLineOptions(self, parser: argparse.ArgumentParser) -> None:
             default=False,
             help="Exit on the first plugin failure.",
         )
+        parser.add_argument(
+            "-l",
+            "--list",
```
makubacki, line 111: Should this be something like `--plugin-list`?

edk2toolext/invocables/edk2_ci_build.py
```diff
@@ -125,8 +134,41 @@ def GetLoggingFileName(self, loggerType: str) -> str:
...
+
+        table = "\n".join([header, separator] + rows)
+
+        print(table)
+
+        print("\nNOTE: You can disable any plugin with PluginName=skip.")
+        print(
+            "NOTE: You can disable all plugins with the --disable-all option, then enable specific plugins with "
+            "<PluginName>=run"
+        )
```
makubacki, line 164: You might also want to consider adding a note that says the plugins are not executed in the order shown.

PR #1007, opened 2025-07-10
Replace az --version call with az version

Nishanth1311:
## Description
az --version finds the latest version information from GitHub (https://github.com/Azure/azure-cli/issues/27437). It can be slow if you have a bad network connection. You may use az version as an alternative as this command doesn't make any network request.

```
>>az --version
azure-cli                         2.75.0

core                              2.75.0
telemetry                          1.1.0

Extensions:
azure-devops                       1.0.1

Dependencies:
msal                            1.33.0b1
azure-mgmt-resource               23.3.0

>az version
{
  "azure-cli": "2.75.0",
  "azure-cli-core": "2.75.0",
  "azure-cli-telemetry": "1.1.0",
  "extensions": {
    "azure-devops": "1.0.1"
  }
}
```

## How This Was Tested
UEFI builds succeed and the az version check passes

### DIFF
--- a/edk2toolext/environment/extdeptypes/az_cli_universal_dependency.py
+++ b/edk2toolext/environment/extdeptypes/az_cli_universal_dependency.py
@@ -58,7 +58,7 @@ class AzureCliUniversalDependency(ExternalDependency):
         if cls.VersionLogged:
             return
         results = StringIO()
-        RunCmd("az", "--version", outstream=results, raise_exception_on_nonzero=True)
+        RunCmd("az", "version", outstream=results, raise_exception_on_nonzero=True)
         results.seek(0)
 
         to_find = ["azure-cli", "azure-devops"]  # find these keys in the version output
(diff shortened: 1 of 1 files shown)


edk2toolext/environment/extdeptypes/az_cli_universal_dependency.py
```diff
@@ -58,7 +58,7 @@ def VerifyToolDependencies(cls: "AzureCliUniversalDependency") -> None:
         if cls.VersionLogged:
             return
         results = StringIO()
-        RunCmd("az", "--version", outstream=results, raise_exception_on_nonzero=True)
+        RunCmd("az", "version", outstream=results, raise_exception_on_nonzero=True)
```
Javagedes, line 61: With the format changing, simply parsing, line by line, the returned versions does not matter.

The reported version for azure-cli is now `'"2.54.0",'` instead of `2.54.0` (A set of double quotes and a comma exist when they should not). and for `azure-devops` the returned value is `"1.0.0"` instead of `1.0.0`. Can you look into fixing this?

I would suggest taking the input and running it through `data = json.loads(results)` and then looking for `azure-cli` and `azure-devops` in the created dictionary.

Javagedes (changes requested):

edk2toolext/environment/extdeptypes/az_cli_universal_dependency.py
Javagedes: @Nishanth1311 bump on this PR comment so we can get it updated and merged in.

PR #1050, opened 2025-12-11
add ability to pass --debug switch to az artifacts download

cmruffin:
Set environment variable AZURE_DEVOPS_EXT_DEBUG=1 to add the --debug switch to az artifacts downloads command.


edk2toolext/environment/extdeptypes/az_cli_universal_dependency.py
```diff
@@ -161,7 +168,7 @@ def _attempt_universal_install(self, install_dir: str) -> None:
             logging.info(f"Failed to parse json data from az command output: {e}")
 
             # Search results to find the json data
-            index_start = results.getvalue().find("{")
+            index_start = results.getvalue().rfind("{")
```
Javagedes: What is the purpose of this particular change? The existing code makes sense to me. use `find` to find the first occurrence of `{` and use `rfind` to find the last occurrence of `}`. This seems like it will break if we have nested curly brackets in the json data output.

